PR #3, opened 2020-07-10
Fix coding standard check

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

Ensure all github actions green.

|    Q          |   A
|-------------- | ------
| QA            | yes

PR #4, opened 2020-07-10
Do not add `.git` since it does not exist when downstream consumers use the action

asgrim:
… the GH Action

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,6 @@ ADD composer.lock /app/composer.lock
 
 RUN COMPOSER_CACHE_DIR=/dev/null composer install --no-dev --no-autoloader
 
-ADD .git /app/.git
 ADD bin /app/bin
 ADD src /app/src
 


asgrim:
CS fix is in #3 , unrelated failure

Dockerfile
```diff
@@ -30,7 +30,6 @@ ADD composer.lock /app/composer.lock
 
 RUN COMPOSER_CACHE_DIR=/dev/null composer install --no-dev --no-autoloader
 
-ADD .git /app/.git
```
Ocramius, line 33: Note: this is needed because of:

https://github.com/laminas/automatic-releases/blob/ea0062141bd984f60e4364600570487cec91b0d2/bin/console.php#L68

Without the git directory, the `ocramius/package-versions` plugin won't detect the correct version to display, which is quite interesting when people use `laminas/automatic-releases@v1` and want to know if something changed in the image.

I think the actual solution is to start publishing the docker images.

Ocramius (changes requested):

Dockerfile
asgrim: Ah stupid `ocramius/package-versions` library TUT

PR #12, opened 2020-07-19
#11 - create+switch default branch on release

### LINKED ISSUE #11: Add mechanism to automatically switch active branch on a new minor or major release
As discussed in the @laminas/technical-steering-committee meeting, we want to move away from the `master` branch naming, towards a more user-friendly naming scheme that exposes version names.

The idea is that whenever a `x.y.0` or `x.0.0` release is tagged, the tool will automatically switch to that release's branch, if it is the latest.

### New minor

For example, given following branches:

 * `1.0.x`
 * `1.1.x` marked as **default** 

When `1.2.0` is tagged, we should see:

 * `1.0.x`
 * `1.1.x`
 * `1.2.x` marked as **default** 

### New major

For example, given following branches:

 * `1.0.x`
 * `1.1.x`
 * `1.2.x` marked as **default** 

When `2.0.0` is tagged, we should see:

 * `1.0.x`
 * `1.1.x`
 * `1.2.x`
 * `2.0.x` marked as **default**

Ocramius:
Fixes #11 by providing default logic for switching to a new default branch at
every release. Specifically:

 * when a newer release target branch exists, use that (i.e. release `1.2.3` => existing branch `1.5.x`,
   switch to `1.5.x`)
 * when a newer release branch does not exist, create it (i.e. release `1.2.3` => existing branch `1.2.x`,
   create `1.3.x` from `1.2.x`, use `1.3.x`)
 * when no release branches exist, skip entire execution

TODOs:

 * [x] try this out in a sandbox repository
 * [x] better documentation? (new `feature/` files)
 * [x] update examples
 * [x] get rid of special handling for `master`, which is not interesting to us anymore
 * [x] document token requirements

Note: it would be good to get this into `1.0.0`, since it breaks BC by changing semantics around the `master` branch naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ to your project or organisation:
  * `GIT_AUTHOR_EMAIL` - email address of the author of your releases: can be an email address of a bot account.
  * `SIGNING_SECRET_KEY` - a **password-less** private GPG key in ASCII format, to be used for signing your releases:
    please use a dedicated GPG subkey for this purpose. Unsigned releases are not supported, and won't be supported.
+ * `ORGANIZATION_ADMIN_TOKEN` - if you use the file from [`examples/.github/workflows/release-on-milestone-closed.yml`](examples/.github/workflows/release-on-milestone-closed.yml),
+   then you have to provide a `ORGANIZATION_ADMIN_TOKEN`, which is a github token with administrative rights over
+   your organisation. This is required for the `laminas:automatic-releases:switch-default-branch-to-next-minor`
+   command, because [changing default branch of a repository currently requires administrative token rights](https://developer.github.com/v3/repos/#update-a-repository).
+   You can generate a token from your [personal access tokens page](https://github.com/settings/tokens/new).
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ to your project or organisation:
    please use a dedicated GPG subkey for this purpose. Unsigned releases are not supported, and won't be supported.
  * `ORGANIZATION_ADMIN_TOKEN` - if you use the file from [`examples/.github/workflows/release-on-milestone-closed.yml`](examples/.github/workflows/release-on-milestone-closed.yml),
    then you have to provide a `ORGANIZATION_ADMIN_TOKEN`, which is a github token with administrative rights over
-   your organisation. This is required for the `laminas:automatic-releases:switch-default-branch-to-next-minor`
+   your organisation (issued by a user that has administrative rights over your project).
+   This is required for the `laminas:automatic-releases:switch-default-branch-to-next-minor`
    command, because [changing default branch of a repository currently requires administrative token rights](https://developer.github.com/v3/repos/#update-a-repository).
    You can generate a token from your [personal access tokens page](https://github.com/settings/tokens/new).
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ this action will perform all following steps (or stop with an error):
  5. create a tag named `x.y.z` on the selected branch, with the generated changelog
  6. publish a release named `x.y.z`, with the generated tag and changelog
  7. create (if applicable), a pull request from the selected branch to the next release branch
+ 8. create (if necessary) a "next minor" release branch `x.y+1.z`
+ 9. switch default repository branch to newest release branch
 
 Please read the [`feature/`](./feature) specification for more detailed scenarios on how the tool is supposed
 to operate.
--- a/examples/.github/workflows/release-on-milestone-closed.yml
+++ b/examples/.github/workflows/release-on-milestone-closed.yml
@@ -17,7 +17,7 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Release"
-        uses: "laminas/automatic-releases@v1"
+        uses: "laminas/automatic-releases@v1.0.0"
         with:
           command-name: "laminas:automatic-releases:release"
         env:
@@ -27,7 +27,7 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Create Merge-Up Pull Request"
-        uses: "laminas/automatic-releases@v1"
+        uses: "laminas/automatic-releases@v1.0.0"
         with:
           command-name: "laminas:automatic-releases:create-merge-up-pull-request"
         env:
@@ -37,11 +37,11 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Create and/or Switch to new Release Branch"
-        uses: "laminas/automatic-releases@v1"
+        uses: "laminas/automatic-releases@v1.0.0"
         with:
           command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
         env:
-          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/examples/.github/workflows/release-on-milestone-closed.yml
+++ b/examples/.github/workflows/release-on-milestone-closed.yml
@@ -35,3 +35,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create and/or Switch to new Release Branch"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/feature/automated-releases.feature
+++ b/feature/automated-releases.feature
@@ -51,7 +51,9 @@ Feature: Automated releases
     When I close milestone "1.1.0"
     Then the tool should have halted with an error
 
-  Scenario: If a minor release branch exists, the tool creates a new minor release from there
+  Scenario: If a minor release branch exists, when closing the minor release milestone,
+  the tool tags the minor release from the branch, and creates a pull request
+  against the next newer minor release branch.
     Given following existing branches:
       | name   |
       | 1.1.x  |
--- a/feature/automated-releases.feature
+++ b/feature/automated-releases.feature
@@ -3,8 +3,10 @@ Feature: Automated releases
 
   Scenario: If no major release branch exists, the tool should not create a new major release
     Given following existing branches:
-      | name  |
-      | 1.0.x |
+      | name    |
+      | 1.0.x   |
+      | master  |
+      | develop |
     And following open milestones:
       | name  |
       | 2.0.0 |
@@ -13,8 +15,10 @@ Feature: Automated releases
 
   Scenario: If no major release branch exists, the tool should not create a new minor release
     Given following existing branches:
-      | name  |
-      | 1.0.x |
+      | name    |
+      | 1.0.x   |
+      | master  |
+      | develop |
     And following open milestones:
       | name  |
       | 1.1.0 |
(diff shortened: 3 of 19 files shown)


src/Git/Value/MergeTargetCandidateBranches.php
```diff
@@ -101,4 +102,41 @@ public function branchToMergeUp(SemVerVersion $version): ?BranchName
             ? null
             : $branch;
     }
+
+    /** @TODO move `master` exclusion into ctor - `master` is no longer a valid release branch */
```
Ocramius: This change will likely lead to a BC break. Not a problem, since the library is not yet released, but good to keep in mind that the tool will change massively, and that `feature/` files will need adjusting

src/Git/Value/MergeTargetCandidateBranches.php
```diff
@@ -101,4 +102,41 @@ public function branchToMergeUp(SemVerVersion $version): ?BranchName
...
+    public function newestReleaseBranch(): ?BranchName
+    {
+        return array_values(array_filter(
+            array_reverse($this->sortedBranches),
+            static fn (BranchName $branch): bool => $branch->name() !== 'master'
+        ))[0] ?? null;
+    }
+
+    /** @TODO move `master` exclusion into ctor - `master` is no longer a valid release branch */
```
Ocramius: Same as https://github.com/laminas/automatic-releases/pull/12/files#r456971170

Ocramius:
Trying this out in the real world, it seems that my github actions `GITHUB_TOKEN` doesn't suffice for this sort of interaction, since that token cannot be used to `PATCH /repos/:org/:name`

Also tried to use a custom token, but no luck: pinging github support.

Ocramius:
Yep, seems like `GITHUB_TOKEN` is scoped too strictly for this: `PATCH /repos/:org/:name` is not possible with that, and instead a custom admin-alike token is needed.

Don't know yet what the precise scope for it is.

Using a custom token produced the expected results at https://github.com/Ocramius/just-trying-automated-releases/runs/888025697?check_suite_focus=true, in which, by releasing `1.1.0`, branch `1.2.x` was created, published, and the repository switched to it.

Ocramius:
Confirmed with github support:

 > The `GITHUB_TOKEN` is limited these permissions in the repository containing the workflow:
 > 
 > https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token
 > 
 > The following page details which GitHub App permissions are required for each endpoint and shows that making changes to the repository's settings, including changing the default branch, requires `write` on `administration`:
 > 
 > https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-administration
 > 
 > If you create a personal access token, the token will need the `repo` permission and will need to belong to a user with `admin` permissions on the repository.

src/Application/Command/SwitchDefaultBranchToNextMinor.php
```diff
@@ -0,0 +1,82 @@
...
+        );
+
+        $mergeCandidates = $this->getMergeCandidates->__invoke($repositoryPath);
+        $releaseVersion  = $event->version();
+        $newestBranch    = $mergeCandidates->newestReleaseBranch();
+
+        if ($newestBranch === null) {
+            $output->writeln('No stable branches found: cannot switch default branch');
+
+            return 0;
```
rieschl, line 65: shouldn't that return a non-zero code? also, there are `Command::SUCCESS` and `Command::FAILURE` for better readability.

src/Github/Api/V3/SetDefaultBranchThroughApiCall.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->apiToken       = $apiToken;
+    }
+
+    public function __invoke(
+        RepositoryName $repository,
+        BranchName $defaultBranch
+    ): void {
+        $request = $this->messageFactory
+            ->createRequest(
+                'PATCH',
```
rieschl, line 43: `\Fig\Http\Message\RequestMethodInterface::METHOD_PATCH`?

src/Application/Command/SwitchDefaultBranchToNextMinor.php
Ocramius: This is not intended to make the build fail: it is true that no stable branches may exist at all (happens when tagging a first release)

src/Github/Api/V3/SetDefaultBranchThroughApiCall.php
Ocramius: Good one! I didn't even know of these constants \o/

Ocramius: Hmm, no, those interfaces don't exist in current dependencies :\

rieschl: it's here https://github.com/php-fig/http-message-util/blob/master/src/RequestMethodInterface.php i.e. in `fig/http-message-util`

Ocramius: Yeh, I would avoid importing a new package just for that

examples/.github/workflows/release-on-milestone-closed.yml
```diff
@@ -35,3 +35,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create and/or Switch to new Release Branch"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
```
Ocramius: As discussed in https://github.com/laminas/automatic-releases/pull/12#issuecomment-660946161, this will need a separate token

Ocramius: This now uses a completely independent environment variable

feature/automated-releases.feature
```diff
@@ -23,39 +27,56 @@ Feature: Automated releases
...
+      | 1.0.x   |
+      | 2.0.x   |
+      | master  |
+      | develop |
     And following open milestones:
       | name  |
       | 2.0.0 |
     When I close milestone "2.0.0"
     Then tag "2.0.0" should have been created on branch "master"
     And branch "2.0.x" should have been created from "master"
```
weierophinney: This scenario confuses me:

- Why would the 2.0.0 tag be created from the `master` branch?
- Why would it create the `2.0.x` branch from `master`, considering the branch already exists?

feature/automated-releases.feature
```diff
@@ -23,39 +27,56 @@ Feature: Automated releases
...
+      | 2.0.x   |
+      | master  |
+      | develop |
     And following open milestones:
       | name  |
       | 1.1.0 |
     When I close milestone "1.1.0"
-    Then tag "1.1.0" should have been created on branch "master"
-    And branch "1.1.x" should have been created from "master"
+    Then the tool should have halted with an error
```
weierophinney, line 52: I'm assuming this is because the `1.1.x` branch does not exist? (just want a clarification)

feature/automated-releases.feature
```diff
@@ -23,39 +27,56 @@ Feature: Automated releases
...
       | 1.1.x  |
+      | 1.2.x  |
       | master |
     And following open milestones:
       | name  |
       | 1.1.0 |
     When I close milestone "1.1.0"
     Then tag "1.1.0" should have been created on branch "1.1.x"
-    And a new pull request from branch "1.1.x" to "master" should have been created
+    And a new pull request from branch "1.1.x" to "1.2.x" should have been created
```
weierophinney: This one feels mis-named. Maybe "If a minor release branch exists, when closing the minor release milestone, the tool tags the minor release from the branch, and creates a pull request against the next newer minor release branch."

feature/automated-releases.feature
```diff
@@ -23,39 +27,56 @@ Feature: Automated releases
...
+
+  Scenario: If a no newer release branch exists, the tool will not create a pull request against it
```
weierophinney:
```suggestion
  Scenario: If no newer release branch exists, the tool will not create any pull requests
```

feature/default-branch-switching.feature
```diff
@@ -0,0 +1,85 @@
...
+    When I close milestone "1.2.0"
+    Then these should be the existing branches:
+      | branch |
+      | 1.0.x  |
+      | 1.1.x  |
+      | 1.2.x  |
+      | 1.3.x  |
+    And the default branch should be "1.3.x"
+
+  Scenario: A pre-existing minor release branch is set as default branch on release
```
weierophinney:
```suggestion
  Scenario: The latest pre-existing minor release branch is set as default branch on a successful release
```

weierophinney (changes requested):
Mainly a few questions.

feature/automated-releases.feature
Ocramius: This is indeed an oversight: I will adjust 👍

feature/automated-releases.feature
Ocramius: Correct: no auto-magic branch switching 👍

feature/automated-releases.feature
Ocramius: It is correct for the tool fo have created a PR from 1.1.x to 1.2.x here though: are you suggesting a change in wording?

src/Github/Api/V3/SetDefaultBranchThroughApiCall.php
```diff
@@ -0,0 +1,69 @@
...
+                self::API_ROOT . 'repos/' . $repository->owner() . '/' . $repository->name()
+            )
+            ->withAddedHeader('Content-Type', 'application/json')
+            ->withAddedHeader('User-Agent', 'Ocramius\'s minimal API V3 client')
+            ->withAddedHeader('Authorization', 'bearer ' . $this->apiToken);
```
asgrim, line 48: Is `Bearer` case sensitive?

README.md
```diff
@@ -25,6 +25,11 @@ to your project or organisation:
  * `GIT_AUTHOR_EMAIL` - email address of the author of your releases: can be an email address of a bot account.
  * `SIGNING_SECRET_KEY` - a **password-less** private GPG key in ASCII format, to be used for signing your releases:
    please use a dedicated GPG subkey for this purpose. Unsigned releases are not supported, and won't be supported.
+ * `ORGANIZATION_ADMIN_TOKEN` - if you use the file from [`examples/.github/workflows/release-on-milestone-closed.yml`](examples/.github/workflows/release-on-milestone-closed.yml),
+   then you have to provide a `ORGANIZATION_ADMIN_TOKEN`, which is a github token with administrative rights over
+   your organisation. This is required for the `laminas:automatic-releases:switch-default-branch-to-next-minor`
+   command, because [changing default branch of a repository currently requires administrative token rights](https://developer.github.com/v3/repos/#update-a-repository).
+   You can generate a token from your [personal access tokens page](https://github.com/settings/tokens/new).
```
asgrim, line 33: The "type" of token is a personal access token, but the naming suggests a different "type" (an "organization admin token"). The link here is the only clue it's a personal access token issued to a user who has organization admin rights. Not sure how to clarify this, but it took me a bit of working out reading the docs given.

src/Application/Command/SwitchDefaultBranchToNextMinor.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->push                = $push;
+        $this->switchDefaultBranch = $switchDefaultBranch;
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $event          = $this->loadGithubEvent->__invoke();
+        $repositoryPath = $this->variables->githubWorkspacePath();
+
+        Assert::directory($repositoryPath . '/.git');
```
asgrim, line 50: Does `symfony/console` catch exceptions and turn them into non-zero exit codes? Or do we need special handling if this assertion fails?

src/Git/Value/BranchName.php
```diff
@@ -45,11 +45,6 @@ public function isReleaseBranch(): bool
         return preg_match('/^(v)?\d+\\.\d+(\\.x)?$/', $this->name) === 1;
     }
 
-    public function isNextMajor(): bool
```
asgrim, line 48: Not `private`, is this a missed BC break? `roave/backward-compatibility-check` could help identify stuff like this, as we already reached `1.*`

src/Git/Value/BranchName.php
```diff
@@ -70,6 +65,13 @@ public function majorAndMinor(): array
         return [$major, $minor];
     }
 
+    public function targetVersion(): SemVerVersion
```
asgrim: Not a fan of this generic naming; my question is "what is the target?" surely the target varies depending on the type of release being done? Just a minor naming thing, admittedly.

src/Github/Api/V3/SetDefaultBranchThroughApiCall.php
Ocramius: Apparently not: tried it manually though

src/Application/Command/SwitchDefaultBranchToNextMinor.php
Ocramius: Yes, exceptions are caught and turned into non-zero exit codes :+1:

src/Git/Value/BranchName.php
Ocramius: Luckily, this library has `"type": "project"`, so it's not meant to be incorporated and relied upon at code level.

What we need to keep stable is the CLI behavior :+1:

Ocramius: > we already reached `1.*`

We're working on `1.0.x`, but we never released BTW

README.md
Ocramius: Added further detail to this wording

src/Git/Value/BranchName.php
Ocramius: Reworded to clarify what this `SemVerVersion` is :)

feature/automated-releases.feature
Ocramius: Adjusted: also, no branch created here: that's done by the `laminas:automatic-releases:switch-default-branch-to-next-minor` command instead.

feature/automated-releases.feature
Ocramius: Got it now: wording adjusted

PR #18, opened 2020-07-31
Allow using existing Keep-A-Changelog files for release notes

### LINKED ISSUE #5: Use CHANGELOG format in releases
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

It was mentioned in chat that we should be using `CHANGELOG` format for tag/release descriptions. This needs more work to be implemented, so mainly opening this as a placeholder for intended future implementation.

weierophinney:
This patch modifies the workings of the `ReleaseCommand` slightly.

It now accepts a new argument `Laminas\AutomaticReleases\Changelog\ReleaseChangelog $releaseChangelog`. The `ReleaseChangelog` interface is intended to prepare a changelog in Keep-A-Changelog format for release (generally by adding the release date), commit it, and push it back to the origin repository.

To accomplish this, I created:

- `CommitFile`, which is intended for, you guessed it, committing a file to the local checkout. One implementation was provided, `CommitFileViaConsole`.
- `Laminas\AutomaticReleases\Changelog\ReleaseChangelogViaKeepAChangelog`, which uses tooling from phly/keep-a-changelog to add the release date to the `CHANGELOG.md` file, and then a `CommitFile` and `Push` instance to commit and push the changes. If the file does not exist, is in a different format, or has no changes necessary, it acts as a no-op.

To allow more ways to generate the release text, I added two additional arguments to the `CreateReleaseText` interface: `BranchName $sourceBranch` and `string $repositoryPath`. Additionally, I added a `canCreateReleaseText(): bool` method that accepts the same arguments as `__invoke()`; this allows testing to see if a given implementation will be able to create the release text.

From there, I added:

- `CreateReleaseTextViaKeepAChangelog`. It's `canCreateReleaseText()` method returns `false` if the file does not exist, is malformed, or does not have an entry corresponding to the version. Otherwise, the `__invoke()` method extracts the changelog entry using the phly/keep-a-changelog `ChangelogParser`.
- `AppendingCreateReleaseTextAggregate`, which accepts a list of `CreateReleaseText` instances. It filters for those capable of creating release text, and then concatenates them with `\n\n-----\n\n`.

When creating the `ReleaseCommand` in `bin/console.php`, I now create an `AppendingCreateReleaseTextAggregate` that uses a `CreateReleaseTextViaKeepAChangelog` instance and a `CreateReleaseTextThroughChangelog` instance.

Fixes #5


composer.json
```diff
@@ -10,9 +10,10 @@
     },
     "require": {
         "php": "^7.4.7",
+        "composer/package-versions-deprecated": "^1.10.99",
```
Ocramius, line 13: Note: we're already on `^7.4.7`, so a dependency change shouldn't be needed here

src/Git/CommitFile.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Laminas\AutomaticReleases\Git;
+
+interface CommitFile
+{
+    public function __invoke(
+        string $repositoryDirectory,
+        string $filename,
+        string $commitMessage,
+        string $author
```
Ocramius: This shouldn't be needed, in theory, given that the repository is pre-configured with an author email and name?

Ocramius: Note also that git will honor any environment variables regarding current author/email

src/Changelog/ReleaseChangelogAndFetchContents.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Changelog;
+
+interface ReleaseChangelogAndFetchContents
```
Ocramius: The purpose of this interface seems to be unclear and/or overloaded.

I think we should instead:

 1. generate the changelog text
 2. replace it in the pre-existing `CHANGELOG.md` if applicable

Those should be two steps, where one is a pure read operation, and the other is a read+write operation

src/Changelog/ReleaseChangelogEvent.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Laminas\AutomaticReleases\Changelog;
+
+use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetMilestoneChangelog\Response\Milestone;
+use Laminas\AutomaticReleases\Github\Value\RepositoryName;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class ReleaseChangelogEvent
```
Ocramius: This should be marked with `/** @psalm-immutable */`

In addition to that, `InputInterface` and `OutputInterface` should not be part of it, as they are part of the application layer of this system, and mutable objects too.

Overall, this object seems to contain too much state that obfuscates the real dependency of consumers of it: splitting it into parameters will probably highlight some areas with too much complexity in them.

src/Changelog/UseKeepAChangelogEventsToReleaseAndFetchChangelog.php
```diff
@@ -0,0 +1,73 @@
...
+use Laminas\AutomaticReleases\Git\Push;
+use Phly\KeepAChangelog\Version\ReadyLatestChangelogEvent;
+use Psr\EventDispatcher\EventDispatcherInterface;
+
+use function assert;
+use function date;
+use function file_exists;
+use function sprintf;
+
+class UseKeepAChangelogEventsToReleaseAndFetchChangelog implements
```
Ocramius: The `CreateChangelogViaMilestone` and `UseKeepAChangelogEventsToReleaseAndFetchChangelog` don't really do the same things:

 * one creates a commit on the repo (this class)
 * the other (`UseKeepAChangelogEventsToReleaseAndFetchChangelog`) produces text to be used for the release

They should not share the same interface, and should not be put together under `ReleaseChangelogAndFetchContentsAggregate`.

In fact, the nullability in `    public function __invoke(ReleaseChangelogEvent $releaseChangelogEvent): ?string` indicates profound differences in behavior, and force you to implement a fallback mechanism if generating a changelog failed.

I think we should:

 1. use keep a changelog to generate `CHANGELOG.md` (and we do use it explicitly before creating a tag)
 2. use `CreateChangelogViaMilestone` to generate the tag description and release notes (same as it was before, relatively UI friendly for github, and still references a commit that contains a `CHANGELOG.md` diff)

Alternatively, we can either:

 * append `CreateChangelogViaMilestone` to tag description/release name, in addition to what `UseKeepAChangelogEventsToReleaseAndFetchChangelog` generated
 * scrap `CreateChangelogViaMilestone`, and generate everything with `UseKeepAChangelogEventsToReleaseAndFetchChangelog`

This means reverting some of these abstractions, and calling the logic sequentially:

```php
$keepAChangelogText = ($this->keepAChangelog)(...);
$releaseText = ($this->createChangelogViaMilestone)(...);

$tag(..., $releaseText . "\n\n" . $keepAChangelogText);
```

No need to aggregate these disjunct pieces of logic together, and no need to change pre-existing tests (besides CLI command tests)

Ocramius (changes requested):

composer.json
weierophinney: It's needed because phly/keep-a-changelog uses it; otherwise, there are conflicts.

src/Git/CommitFile.php
weierophinney: I was unclear if this were the case, but seeing how the `CreateTagViaConsole` class is written, it must be. I'll revert that in my next revisions.

composer.json
boesing: Yeah, something messed up with that replacement. Had some weird problems when using laminas-cli aswell...
Not sure what causes those conflicts...

src/Changelog/UseKeepAChangelogEventsToReleaseAndFetchChangelog.php
weierophinney: I disagree with some of this assessment.

KeepAChangelog does not **generate** a changelog; it **is** the changelog. The phly/keep-a-changelog tooling was being used to:

- Set the release date for the given changelog version.
- Retrieve the contents for that changelog version for purposes of the GitHub release.

That first part is where we have an overlap of responsibilities, as it also involves creating a commit.

The reason I want the Keep-A-Changelog format in the GitHub release is because if you search for `{package} {version} changelog`, you end up at the GitHub releases page, which is where you'll then read what _changed_ for a given version. The human-readable format is important here.

That said, I really like what JwageGenerateChangelog creates, as it provides _attribution_, which is also something I think it's important for us to do as a community OSS project.

On top of all that: I can see value in this tooling outside of Laminas. Yes, we're the primary consumers right now, but you've done a great job of engineering it such that it is generalized enough to work with any project that wants to adopt the workflow it implements for tagging, releasing, and up-merging. As such, we cannot necessarily assume that a given project has a changelog, much less in keep-a-changelog format.

So, here's my idea for moving forward:

- I'll split out the "add date to changelog and commit file" aspect into a separate utility that is passed to `ReleaseCommand` and invoked prior to retrieving the changelog text. This utility will be a no-op if no changelog file is present or if the changelog is not in keep-a-changelog format.

- I'll merge the `ReleaseChangelogAndFetchContents` into `CreateReleaseText` interfaces, adding arguments for source branch and repository directory to the latter, and make the current `UseKeepAChangelog....` implement that instead, having it return an empty string if no keep-a-changelog file is found. I'll also create an "appending aggregate" version that executes each and concatenates it to the previous.

I think this will resolve the architectural issues you've flagged.

Ocramius: > KeepAChangelog does not generate a changelog; it is the changelog. The phly/keep-a-changelog tooling was being used to:

One is `f :: API -> Text`, the other is `f :: API -> (Text, GitCommit)` - very different underlying architecture

Ocramius: > I think this will resolve the architectural issues you've flagged.

Will likely be able to do another review tonight :)

Ocramius: Discussed in chat. It is clear now that this should be a 2 step operation of: 

1. generating the text (to be manipulated and/or used directly)
2. committing it to repo (separate service/function that does just that)

weierophinney:
It looks like the failing unit test is likely due to a missing git author/email, but it's hard to determine. Tests pass fine locally, which is why I'm inferring it's an environment issue.

weierophinney:
> It looks like the failing unit test is likely due to a missing git author/email, but it's hard to determine.

Verified: https://github.com/laminas/automatic-releases/pull/18/checks?check_run_id=946634877#step:8:68

Do you want me to add the git author/email to the travis env as part of this patch?

weierophinney:
@Ocramius Okay, ready for another round of review. :smile:

test/unit/Changelog/ReleaseChangelogViaKeepAChangelogTest.php
```diff
@@ -0,0 +1,212 @@
...
+    private const INVALID_CHANGELOG = <<< 'END'
+        # NOT A CHANGELOG
+
+        This file is not a changelog.
+
+        ## Bad headers
+
+        It contains bad headers, among other things.
+
+        END;
```
weierophinney, line 159: Since we're on 7.4... One cool feature of 7.3 is that it will now strip indentation from heredoc/nowdoc contents if the end marker is indented. Makes these constant declarations far more readable!

test/unit/Github/CreateReleaseTextViaKeepAChangelogTest.php
```diff
@@ -0,0 +1,204 @@
...
+            - Nothing.
+            
+            ### Removed
+            
+            - Nothing.
+            
+            ### Fixed
+            
+            - Nothing.
+            END, $date);
```
weierophinney, line 112: Also, heredoc/nowdoc no longer require the end marker to be on a line of their own followed by a `;`, which makes this work.

src/Github/AppendingCreateReleaseTextAggregate.php
```diff
@@ -0,0 +1,77 @@
...
+        string $repositoryDirectory
+    ): string {
+        $items = array_map(
+            static fn (CreateReleaseText $strategy): string => $strategy($milestone, $repositoryName, $semVerVersion, $sourceBranch, $repositoryDirectory),
+            array_filter(
+                $this->strategies,
+                // phpcs:disable
+                fn(CreateReleaseText $strategy): bool => $strategy->canCreateReleaseText($milestone, $repositoryName, $semVerVersion, $sourceBranch, $repositoryDirectory)
+                // phpcs:enable
+            )
```
weierophinney: I used arrow functions here so I wouldn't need to use `use(...)` with that long list of arguments.

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
```diff
@@ -0,0 +1,141 @@
...
+    private function createReadyLatestChangelogEvent(
+        string $changelogFile,
+        string $versionString
+    ): ReadyLatestChangelogEvent {
+        /**
+         * Hard-coded extension to allow setting version to known value
+         *
+         * @psalm-suppress PropertyNotSetInConstructor
+         */
+        $event = new class ($versionString) extends ReadyLatestChangelogEvent {
```
weierophinney: I used anonymous classes in this implementation as they are one-offs, and only set the values necessary for the specific event listeners I'm manually triggering. If you want me to extract these, let me know.

weierophinney (commented):
A few comments on implementation details...

bin/console.php
```diff
@@ -56,17 +60,24 @@ static function (int $errorCode, string $message = '', string $file = '', int $l
...
-    $createReleaseText    = new CreateReleaseTextThroughChangelog(JwageGenerateChangelog::create(
+    $commit               = new CommitFileViaConsole();
+    $push                 = new PushViaConsole();
+    $releaseChangelog     = new ReleaseChangelogViaKeepAChangelog($commit, $push);
+    $createCommitText     = new CreateReleaseTextThroughChangelog(JwageGenerateChangelog::create(
         $makeRequests,
         $httpClient
     ));
-    $push                 = new PushViaConsole();
+    $createReleaseText    = new AppendingCreateReleaseTextAggregate([
```
Ocramius: The name `AppendingCreateReleaseTextAggregate` is quite terrible - can we synthesise into a small sentence?

composer.json
Ocramius: Could be any of the CLI tools - not a big deal though, as we're free to use `composer/composer:^2` here (it's an application, not a library)

src/Changelog/ReleaseChangelog.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Changelog;
+
+use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+
+interface ReleaseChangelog
```
Ocramius: This name is not clear: is it  supposed to be "commit release changelog" perhaps? Since its signature is `: void`, I suppose it **does** something

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
```diff
@@ -0,0 +1,141 @@
...
+use Laminas\AutomaticReleases\Git\Push;
+use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+use Phly\KeepAChangelog\Common\DiscoverChangelogEntryListener;
+use Phly\KeepAChangelog\Config;
+use Phly\KeepAChangelog\Version\ReadyLatestChangelogEvent;
+use Phly\KeepAChangelog\Version\SetDateForChangelogReleaseListener;
+use Symfony\Component\Console\Output\NullOutput;
+
+use function date;
```
Ocramius: This should not be: if the current date/time is needed, let's inject it via a `Clock` (https://github.com/lcobucci/clock is a simple abstraction for it)

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
```diff
@@ -0,0 +1,141 @@
...
+        if (! $this->updateChangelog($changelogFile, $versionString)) {
+            // Failure to update; nothing to commit
+            return;
+        }
+
+        ($this->commitFile)(
+            $repositoryDirectory,
+            $sourceBranch,
+            'CHANGELOG.md',
+            sprintf('%s readiness', $versionString)
```
Ocramius: `1.2.3 readiness` doesn't sound like an awesome commit text: can we use something that explains that we did just provide a changelog text?

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
```diff
@@ -0,0 +1,141 @@
...
+        );
+
+        ($this->push)($repositoryDirectory, $sourceBranch->name());
+    }
+
+    private function updateChangelog(string $changelogFile, string $versionString): bool
+    {
+        $event = $this->createReadyLatestChangelogEvent($changelogFile, $versionString);
+
+        (new DiscoverChangelogEntryListener())($event);
```
Ocramius: These listeners are bizarre: I know they are mostly imported, and you mostly did C&P of some upstream code: still feels weird for them to mutate `$event`.

Probably design similar to what we had in `laminas/laminas-mvc`?

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
Ocramius: I think this is an OK trade-off: we can't go and change the whole upstream library for this :)

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * @psalm-param non-empty-string $repositoryDirectory
+     */
+    public function __invoke(
+        string $repositoryDirectory,
+        SemVerVersion $version,
+        BranchName $sourceBranch
+    ): void {
+        $changelogFile = $repositoryDirectory . '/CHANGELOG.md';
```
Ocramius: `CHANGELOG.md` is referenced twice in here: in order to prevent typos in future, let's move it to a `private const`

src/Github/AppendingCreateReleaseTextAggregate.php
```diff
@@ -0,0 +1,77 @@
...
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetMilestoneChangelog\Response\Milestone;
+use Laminas\AutomaticReleases\Github\Value\RepositoryName;
+use Webmozart\Assert\Assert;
+
+use function array_filter;
+use function array_map;
+use function implode;
+
+final class AppendingCreateReleaseTextAggregate implements CreateReleaseText
```
Ocramius: Mentioned above: the naming is a bit weird, but that's because it follows the pattern rather than the actual intent.

Something like `ConcatenateMultipleReleaseTexts` would suffice

src/Github/AppendingCreateReleaseTextAggregate.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    private const CONCATENATION_STRING = "\n\n-----\n\n";
+
+    /** @var CreateReleaseText[] */
+    private array $strategies;
+
+    /** @param CreateReleaseText[] $strategies */
```
Ocramius: `/** @psalm-param non-empty-list<CreateReleaseText> */` and the assertion goes away :D

src/Github/AppendingCreateReleaseTextAggregate.php
Ocramius: Not sure why phpcs complains above though :thinking: 

Perhaps it wants `static fn`?

src/Github/AppendingCreateReleaseTextAggregate.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __construct(array $strategies)
+    {
+        Assert::notEmpty($strategies);
+        $this->strategies = $strategies;
```
Ocramius: More than `strategies` (pattern name), possibly `$this->releaseTextGenerators` or such?

src/Github/AppendingCreateReleaseTextAggregate.php
```diff
@@ -0,0 +1,77 @@
...
+            static fn (CreateReleaseText $strategy): string => $strategy($milestone, $repositoryName, $semVerVersion, $sourceBranch, $repositoryDirectory),
+            array_filter(
+                $this->strategies,
+                // phpcs:disable
+                fn(CreateReleaseText $strategy): bool => $strategy->canCreateReleaseText($milestone, $repositoryName, $semVerVersion, $sourceBranch, $repositoryDirectory)
+                // phpcs:enable
+            )
+        );
+
+        /** @psalm-var non-empty-string $changelog */
```
Ocramius: Considering that (in theory) no generator may support this, a non-empty-string is not guaranteed.

In order to prevent that, let's add `Assert::notEmpty()` **after** `$changelog` has been assigned.

The inline docblock is coding a too heavy assumption here, so we turn it into an explicit crash in case of misconfiguration.

src/Github/CreateReleaseTextThroughChangelog.php
```diff
@@ -56,6 +59,16 @@ public function __invoke(
         return $text;
     }
 
+    public function canCreateReleaseText(
+        Milestone $milestone,
+        RepositoryName $repositoryName,
+        SemVerVersion $semVerVersion,
+        BranchName $sourceBranch,
+        string $repositoryDirectory
```
Ocramius, line 67: Not important in this patch, so feel free to dismiss (or creating a new issue for me to solve later), but I think we're repeating `$repositoryDirectory` **a lot** across this codebase.

A value type would prevent us from passing a random `non-empty-string` in here.

src/Github/CreateReleaseTextViaKeepAChangelog.php
```diff
@@ -0,0 +1,60 @@
...
+class CreateReleaseTextViaKeepAChangelog implements CreateReleaseText
+{
+    public function __invoke(
+        Milestone $milestone,
+        RepositoryName $repositoryName,
+        SemVerVersion $semVerVersion,
+        BranchName $sourceBranch,
+        string $repositoryDirectory
+    ): string {
+        /** @psalm-var non-empty-string $changelog */
```
Ocramius: This is not guaranteed by the `ChangelogParser`: let's instead use `Assert::notEmpty()` **after** `$changelog` has been assigned a `string`

test/unit/Changelog/ReleaseChangelogViaKeepAChangelogTest.php
Ocramius: I would still endorse using a different start/end marker depending on context. For example, here I would have used `<<< 'MARKDOWN'`

Not important, since we're in the middle of a relatively clear test

test/unit/Changelog/ReleaseChangelogViaKeepAChangelogTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+        file_put_contents(
+            sprintf('%s/%s', $repo, $filename),
+            $template
+        );
+
+        return $repo;
+    }
+
+    private const INVALID_CHANGELOG = <<< 'END'
```
Ocramius, line 150: What happens if `CHANGELOG.md` is empty?

test/unit/Git/CommitFileViaConsoleTest.php
```diff
@@ -0,0 +1,71 @@
...
+        Assert::notEmpty($checkout);
+
+        $this->checkout = $checkout;
+        unlink($this->checkout);
+
+        (new Process(['git', 'init', $this->checkout]))
+            ->mustRun();
+
+        (new Process(
+            ['git', 'symbolic-ref', 'HEAD', 'refs/heads/1.0.x'],
```
Ocramius, line 42: Fancy trick! This is basically a `git switch`, but to a new branch?

test/unit/Github/CreateReleaseTextViaKeepAChangelogTest.php
Ocramius: Oooh, didn't know! Awesome, but it does indeed confuse the heck out of the github PHP parsing :D

Ocramius (approved):
Giving my early LGTM here.

The remaining comments are:

 * [x] naming (ZF habits are hard to let go of :D )
 * [x] testing (in CI, the `GIT_*` environment variables are not defined, so all mock/stub repos need to be configured with user/email before committing)
 * [x] some minor adjustments around assertions
 * [x] did you try it out manually somewhere?

bin/console.php
weierophinney: How about `AppendSeveralReleaseTexts`?

src/Changelog/ReleaseChangelog.php
weierophinney: Yeah, `CommitReleaseChangelog` makes sense - I was thinking "release" as a verb here, but we also have a concept of "a release" throughout. If that name makes sense, I'll update.

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
weierophinney: As in, inject it as an argument? as a property injected or created via constructor? Creation of the instance would be something like `new FrozenClock(new DateTimeImmutable('now'))` in the `bin/console.php` script?

Essentially, I just need the date string, so internally, where I have `date('Y-m-d')`, this would be come `$clock->now()->format('Y-m-d')`?

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
weierophinney: Will do!

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
Ocramius: in here, `$this->clock->now()->format('Y-m-d')`

In the constructor call, `new SystemClock()` is given as a dependency

bin/console.php
Ocramius: s/append/concatenate, but otherwise  :+1:

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
weierophinney: PSR-14, actually, via the `StoppableEventInterface`. Essentially, dispatchers check for `$event->isPropagationStopped()` after triggering each listener; if it returns true, it doesn't iterate to the next listener. I use this heavily in keep-a-changelog, as it makes it easier for me to report error conditions back to the user and/or to stop doing work if nothing else needs to be done. I use a `failed()` method to check for it in my library, and `isPropagationStopped()` typically checks for that (sometimes it checks other things as well).

src/Changelog/ReleaseChangelogViaKeepAChangelog.php
weierophinney: Will do.

src/Github/CreateReleaseTextThroughChangelog.php
weierophinney: I'd be willing to add that as part of this patch if you'd like; I rely on it **heavily**, as I need to work with the filesystem.

src/Github/CreateReleaseTextViaKeepAChangelog.php
weierophinney: :+1:

That said, `ChangelogParser` will raise an exception if no changelog is found and/or it is empty (it's looking for non-empty contents, IIRC). But the assertion makes it bullet-proof.

src/Github/CreateReleaseTextThroughChangelog.php
Ocramius: Let's defer: I'll create an issue for this. Important right now is testing (which I'm doing on my little frankenstein repo @ https://github.com/Ocramius/just-trying-automated-releases ) and then I think we should just ship this :)

test/unit/Changelog/ReleaseChangelogViaKeepAChangelogTest.php
weierophinney: Parser raises an exception.

test/unit/Git/CommitFileViaConsoleTest.php
weierophinney: Yep! Discovered it when I was trying to figure out how to test this yesterday. What it does is change the symbolic name of the HEAD to what you provide; it's a way to rename the current branch, essentially.

bin/console.php
weierophinney: Never mind - I like your suggestion later of `ConcatenateMultipleReleaseTexts`.

.github/workflows/psalm.yml
```diff
@@ -57,3 +57,8 @@ jobs:
 
       - name: "psalm"
         run: "vendor/bin/psalm --shepherd --stats"
+        env:
+            GIT_AUTHOR_NAME: "Laminas Automatic Releases"
```
Ocramius: Eh, no, let's initialize repositories in tests, rather than requiring environment to be set

weierophinney: The error was from running `git init` within a unit test setup, due to missing author/committer in the env...

Ocramius: Aware: the other tests configure a user/email 👍

Ocramius (approved):
LGTM :ship:

PR #37, opened 2020-08-10
Create new milestones automatically

### LINKED ISSUE #26: Create next milestones after a release is prepared
### Feature Request

See https://github.com/doctrine/automatic-releases/issues/26

After a release is done, we want to implicitly generate new milestones.

```gherkin
Given following milestones:
  | name |
  | 1.0.0 |
When milestone 1.0.0 is closed
Then following milestones should exist:
  | name |
  | 1.0.1 |
  | 1.1.0 |
  | 2.0.0 |
```

This should probably be done as a new command to be added to the automatic-release workflow example.

Note: the command must gracefully ignore pre-existing milestones with same names, and duplicate milestone names must never be possible!

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

geerteltink:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Create next milestones after a release is prepared.

Closes #26

### DIFF
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -55,3 +55,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create new milestones"
+        uses: "./"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -61,7 +61,7 @@ jobs:
         with:
           command-name: "laminas:automatic-releases:create-milestones"
         env:
-          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ composer-validate:
 	composer validate
 
 static-analysis:
-	vendor/bin/phpstan analyse
 	vendor/bin/psalm
 
 test:
--- a/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
+++ b/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
@@ -65,7 +65,7 @@ jobs:
         with:
           command-name: "laminas:automatic-releases:create-milestones"
         env:
-          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
+++ b/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
@@ -59,3 +59,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create new milestones"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/examples/.github/workflows/release-on-milestone-closed.yml
+++ b/examples/.github/workflows/release-on-milestone-closed.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           command-name: "laminas:automatic-releases:create-milestones"
         env:
-          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/examples/.github/workflows/release-on-milestone-closed.yml
+++ b/examples/.github/workflows/release-on-milestone-closed.yml
@@ -52,6 +52,13 @@ jobs:
           command-name: "laminas:automatic-releases:bump-changelog"
         env:
           "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Create new milestones"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/src/Application/Command/CreateMilestones.php
+++ b/src/Application/Command/CreateMilestones.php
@@ -11,6 +11,7 @@ use Laminas\AutomaticReleases\Github\Api\V3\CreateMilestone;
 use Laminas\AutomaticReleases\Github\Api\V3\CreateMilestoneFailed;
 use Laminas\AutomaticReleases\Github\Event\Factory\LoadCurrentGithubEvent;
 use Laminas\AutomaticReleases\Github\Value\RepositoryName;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -20,17 +21,20 @@ final class CreateMilestones extends Command
     private LoadCurrentGithubEvent $loadEvent;
     private GetGithubMilestone $getMilestone;
     private CreateMilestone $createMilestone;
+    private LoggerInterface $logger;
 
     public function __construct(
         LoadCurrentGithubEvent $loadEvent,
         GetGithubMilestone $getMilestone,
-        CreateMilestone $createMilestone
+        CreateMilestone $createMilestone,
+        LoggerInterface $logger
     ) {
         parent::__construct('laminas:automatic-releases:create-milestones');
 
         $this->loadEvent       = $loadEvent;
         $this->getMilestone    = $getMilestone;
         $this->createMilestone = $createMilestone;
+        $this->logger          = $logger;
     }
 
     public function execute(InputInterface $input, OutputInterface $output): int
@@ -54,7 +58,7 @@ final class CreateMilestones extends Command
         try {
             ($this->createMilestone)($repositoryName, $version);
         } catch (CreateMilestoneFailed $e) {
-            $output->writeln($e->getMessage());
+            $this->logger->info($e->getMessage());
             return false;
         }
 
--- a/src/Application/Command/CreateMilestones.php
+++ b/src/Application/Command/CreateMilestones.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Laminas\AutomaticReleases\Application\Command;
 
-use InvalidArgumentException;
 use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
 use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetGithubMilestone;
 use Laminas\AutomaticReleases\Github\Api\V3\CreateMilestone;
@@ -59,6 +58,7 @@ final class CreateMilestones extends Command
             ($this->createMilestone)($repositoryName, $version);
         } catch (CreateMilestoneFailed $e) {
             $this->logger->info($e->getMessage());
+
             return false;
         }
 
--- a/src/Application/Command/CreateMilestones.php
+++ b/src/Application/Command/CreateMilestones.php
@@ -5,12 +5,10 @@ declare(strict_types=1);
 namespace Laminas\AutomaticReleases\Application\Command;
 
 use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
-use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetGithubMilestone;
 use Laminas\AutomaticReleases\Github\Api\V3\CreateMilestone;
 use Laminas\AutomaticReleases\Github\Api\V3\CreateMilestoneFailed;
 use Laminas\AutomaticReleases\Github\Event\Factory\LoadCurrentGithubEvent;
 use Laminas\AutomaticReleases\Github\Value\RepositoryName;
-use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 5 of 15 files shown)


src/Application/Command/CreateMilestones.php
```diff
@@ -0,0 +1,62 @@
...
+
+        return 0;
+    }
+
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): bool
+    {
+        try {
+            ($this->createMilestone)($repositoryName, $version->nextPatch());
+        } catch (InvalidArgumentException $e) {
+            // Log that the milestone exists?
```
geerteltink: I have searched the GitHub API docs and haven't found a way to search for a milestone by its title. I've tried manually to create milestones and they failed for duplicate titles. I'm hoping that the API works the same. This needs to be tested.

src/Git/Value/SemVerVersion.php
```diff
@@ -54,11 +54,21 @@ public function minor(): int
...
+    {
+        return new self($this->major, $this->minor, $this->patch + 1);
+    }
+
     public function nextMinor(): self
     {
         return new self($this->major, $this->minor + 1, 0);
     }
 
+    public function nextMajor(): self
```
geerteltink, line 67: This needs a test.

src/Git/Value/SemVerVersion.php
```diff
@@ -54,11 +54,21 @@ public function minor(): int
         return $this->minor;
     }
 
+    public function nextPatch(): self
```
geerteltink, line 57: This needs a test.

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
```diff
@@ -0,0 +1,79 @@
...
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\UriInterface;
+use Webmozart\Assert\Assert;
+
+use function Safe\json_decode;
+use function Safe\json_encode;
+use function sprintf;
+
+final class CreateMilestoneThroughApiCall implements CreateMilestone
```
geerteltink, line 18: Missing tests.

src/Application/Command/CreateMilestones.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->createMilestoneIfNotExists($repositoryName, $releaseVersion->nextMajor());
+
+        return 0;
+    }
+
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): bool
+    {
+        try {
+            ($this->createMilestone)($repositoryName, $version->nextPatch());
+        } catch (InvalidArgumentException $e) {
```
Ocramius: This exception type is too generic

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
```diff
@@ -0,0 +1,79 @@
...
+                'body'     => sprintf('{"title":"%s"}', $version->fullReleaseName()),
+            ]));
+
+        $response = $this->client->sendRequest($request);
+
+        $responseBody = $response
+            ->getBody()
+            ->__toString();
+
+        Assert::greaterThanEq($response->getStatusCode(), 200);
```
Ocramius: In case of a duplicate, I expect an HTTP 409 here (to be verified): in that case, let's throw a specific exception, if we want to rely on it.

Ocramius (changes requested):

examples/.github/workflows/release-on-milestone-closed.yml
```diff
@@ -37,11 +37,21 @@ jobs:
...
+        uses: "laminas/automatic-releases@1.1.0"
         with:
           command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
         env:
           "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create new milestones"
```
Ocramius, line 56: This needs to land also in `.github`, not just in `examples/.github` :+1:

src/Application/Command/CreateMilestones.php
weierophinney: I can vouch for the API failing when the milestone already exists (I ran into this with keep-a-changelog).

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
weierophinney: I just checked. You get a 422 status, and the contents looks like the following:

```json
{
    "documentation_url": "https://docs.github.com/rest/reference/issues#create-a-milestone",
    "errors": [
        {
            "code": "already_exists",
            "field": "title",
            "resource": "Milestone"
        }
    ],
    "message": "Validation Failed"
}
```

src/Application/Command/CreateMilestones.php
weierophinney: > I have searched the GitHub API docs and haven't found a way to search for a milestone by its title.

I've even looked for graphql queries to do this, but haven't found any. Every single example I've seen has you list milestones, and then do an array reduce to find the one matching the title you want.

geerteltink: I'm not sure if loading all milestones and go through them is a good solution if you have a huge amount of milestones. So yes, if we are sure the API fails, we can go with that.

weierophinney: 100% positive it results in an error, so that's the best and easiest path.

geerteltink: I'm doing it differently. GitHub sends a 201 status code if successful. Otherwise I'll throw an exception and pass through the response body for debugging.

geerteltink:
@Ocramius ready for final review. The test repo I used is here: https://github.com/xtreamwayz/refactored-stroopwafel/milestones. All milestones there are automatically created.

.github/workflows/automatic-release.yml
```diff
@@ -55,3 +55,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create new milestones"
+        uses: "./"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
```
Ocramius, line 62: Discussed this elsewhere with @weierophinney - unsure if we should pack this in a separate step or not 🤔

.github/workflows/automatic-release.yml
```diff
@@ -55,3 +55,13 @@ jobs:
...
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
```
Ocramius: I think the normal token suffices for this interaction

Makefile
```diff
@@ -10,7 +10,6 @@ composer-validate:
 	composer validate
 
 static-analysis:
-	vendor/bin/phpstan analyse
```
Ocramius, line 13: Good catch: I should've used the makefile in the targets 😬

examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
```diff
@@ -59,3 +59,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create new milestones"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
```
Ocramius: Similar to above: doesn't need a special token

examples/.github/workflows/release-on-milestone-closed.yml
```diff
@@ -52,6 +52,13 @@ jobs:
           command-name: "laminas:automatic-releases:bump-changelog"
         env:
           "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Create new milestones"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:create-milestones"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
```
Ocramius: Similar to above: doesn't need special token

src/Application/Command/CreateMilestones.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $milestoneClosedEvent = ($this->loadEvent)();
+        $repositoryName       = $milestoneClosedEvent->repository();
+        $releaseVersion       = $milestoneClosedEvent->version();
+        $milestone            = ($this->getMilestone)($repositoryName, $milestoneClosedEvent->milestoneNumber());
+
+        $milestone->assertAllIssuesAreClosed();
```
Ocramius: Does this matter to us? Creating milestones is non-destructive, while creating releases has a downstream effect. Perhaps we can skip this check completely?

src/Application/Command/CreateMilestones.php
```diff
@@ -0,0 +1,67 @@
...
+
+        return 0;
+    }
+
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version, OutputInterface $output): bool
+    {
+        try {
+            ($this->createMilestone)($repositoryName, $version);
+        } catch (CreateMilestoneFailed $e) {
+            $this->logger->info($e->getMessage());
```
Ocramius: In PSR-3, an exception should be given as a logging context to the logger: `['exception' => $e]` as added parameter, specifically

src/Application/Command/CreateMilestones.php
```diff
@@ -0,0 +1,67 @@
...
+        $milestone->assertAllIssuesAreClosed();
+
+        $this->createMilestoneIfNotExists($repositoryName, $releaseVersion->nextPatch(), $output);
+        $this->createMilestoneIfNotExists($repositoryName, $releaseVersion->nextMinor(), $output);
+        $this->createMilestoneIfNotExists($repositoryName, $releaseVersion->nextMajor(), $output);
+
+        return 0;
+    }
+
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version, OutputInterface $output): bool
```
Ocramius: `$output` is no longer in use, and can be removed

src/Github/Api/V3/CreateMilestone.php
```diff
@@ -0,0 +1,17 @@
...
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+use Laminas\AutomaticReleases\Github\Value\RepositoryName;
+use Psr\Http\Message\UriInterface;
+
+interface CreateMilestone
+{
+    public function __invoke(
+        RepositoryName $repository,
+        SemVerVersion $version
+    ): UriInterface;
```
Ocramius: Do we care about the URI at all?

Ocramius (changes requested):
Minor changes needed. Overall OK with adding this as a new step (although perhaps earlier in the release workflow?)

We kinda need to simplify the setup for end users though, so that they don't need to change their configuration all the time (new ticket, perhaps?)

.github/workflows/automatic-release.yml
weierophinney: I'm going back and forth on this.

Having it as a separate step makes it opt-in, so users can decided if this is something they want in their workflow.

On the flip side, as you mention elsewhere, adding stuff like this means we end up having to go back to repos where we've already enabled the workflow to add this step, which is time consuming and a bit of a pain.

One option might be to define a default meta-command that defines the general Laminas workflow, and internally calls each of the various commands, in order. This way as we add to the default workflow, no changes are necessary.

The one problem with this approach would be that we would either need to differentiate between the standard and the admin token in the `Variables` implementation, or we'd have to decide it's okay to use the admin token for any API calls.

But as we continue to add features, this is going to come up again, so we definitely need to address it somehow.

.github/workflows/automatic-release.yml
weierophinney: Just checked, and @Ocramius is right here; there are no requirements outside of standard repo access (i.e., the owner of the token needs write access on the repo, but that's all).

src/Application/Command/CreateMilestones.php
weierophinney: Agreed with @Ocramius. Additionally, this will likely happen as a late step in the workflow, which means we'll be aborting the workflow if this is not already true.

src/Application/Command/CreateMilestones.php
weierophinney: One other point: in previous usage of the logger, I've been using either the interface name or class name (if not implementing an interface) as a prefix for the message: `[InterfaceOrClassName] `. This makes it far easier to understand **what** is generating the log message.

.github/workflows/automatic-release.yml
geerteltink: If you have this in a separate step, users can configure the features they want, or do you mean merge this with the "Release" step?

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
```diff
@@ -0,0 +1,96 @@
...
+        ));
+
+        $request = $this->messageFactory
+            ->createRequest(
+                'POST',
+                self::API_ROOT . 'repos/' . $repository->owner() . '/' . $repository->name() . '/milestones'
+            )
+            ->withAddedHeader('Content-Type', 'application/json')
+            ->withAddedHeader('User-Agent', 'Ocramius\'s minimal API V3 client')
+            ->withAddedHeader('Authorization', 'token ' . $this->apiToken);
```
weierophinney, line 60: Unrelated to this patch, but perhaps we should have a specialized PSR-17 request factory that adds these headers to the request generated by the factory? Thoughts, @Ocramius ?

.github/workflows/automatic-release.yml
Ocramius: I guess it makes sense to have this as separate step :+1:

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
```diff
@@ -0,0 +1,96 @@
...
+
+        Assert::eq($response->getStatusCode(), 201);
+
+        $responseData = json_decode($responseBody, true);
+        Assert::isMap($responseData);
+        Assert::keyExists($responseData, 'html_url');
+        Assert::stringNotEmpty($responseData['html_url']);
+
+        $this->logger->info(sprintf(
+            'Milestone "%s" created',
```
weierophinney: Prefix with the interface or class name, please.

test/unit/Git/Value/SemVerVersionTest.php
```diff
@@ -158,4 +158,31 @@ public function lessThanEqualProvider(): array
             ['1.0.0', '2.0.0', true],
         ];
     }
+
+    /**
+     * @dataProvider nextVersionsProvider
+     */
+    public function testNextVersions(string $version, string $patch, string $minor, string $major): void
```
weierophinney, line 169: These should be documented as `non-empty-string` via annotation.

test/unit/Git/Value/SemVerVersionTest.php
```diff
@@ -158,4 +158,31 @@ public function lessThanEqualProvider(): array
...
+    {
+        self::assertEquals($patch, SemVerVersion::fromMilestoneName($version)->nextPatch()->fullReleaseName());
+        self::assertEquals($minor, SemVerVersion::fromMilestoneName($version)->nextMinor()->fullReleaseName());
+        self::assertEquals($major, SemVerVersion::fromMilestoneName($version)->nextMajor()->fullReleaseName());
+    }
+
+    /**
+     * @return string[][]
+     *
+     * @psalm-return non-empty-list<array{string, string, string, string}>
```
weierophinney: `s/string/non-empty-string/`

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -0,0 +1,173 @@
...
+            ->with('POST', 'https://api.github.com/repos/foo/bar/milestones')
+            ->willReturn(new Request('https://the-domain.com/the-path'));
+
+        $validResponse = (new Response())->withStatus(201);
+
+        $validResponse->getBody()->write(<<<'JSON'
+{
+    "html_url": "http://another-domain.com/the-pr"
+}
+JSON
```
weierophinney: Indent these, please; you can see in some of the other tests (e.g., the keep-a-changelog related tests) where I've done that. Makes reading easier, and supported since PHP 7.3. You just need to have the end marker indented the same amount as the rest of the lines.

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -0,0 +1,173 @@
...
+                    ],
+                    $request->getHeaders()
+                );
+
+                self::assertJsonStringEqualsJsonString(
+                    <<<'JSON'
+{
+    "title": "1.2.3"
+}
+JSON
```
weierophinney: Same comment here.

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -0,0 +1,173 @@
...
+            $this->httpClient,
+            $this->apiToken,
+            $this->logger
+        );
+    }
+
+    public function testSuccessfulRequest(): void
+    {
+        $this
+            ->messageFactory
```
weierophinney: I tend to do these on the same line, as the following lines _act on the factory object_, not `$this`. But that's a personal preference.

weierophinney (changes requested):

src/Github/Api/V3/CreateMilestoneThroughApiCall.php
Ocramius: @weierophinney I don't think this is the responsibility of the factory, but rather of a Github API client: the fact that a token is added, and that a user-agent is added, is specific to Github's requirements, and we can decorate the request there if needed.

I would keep it as-is unless we keep adding more and more Github API interactions.

Ocramius (approved):
LGTM :ship:

geerteltink:
Running some final tests now.

Ocramius:
@geerteltink ping me and I'll get 1.4.0 through tonight, if all worked :-)

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -0,0 +1,168 @@
...
+                    $request->getHeaders()
+                );
+
+                self::assertJsonStringEqualsJsonString(
+                    <<<'JSON'
+                    {
+                        "title": "1.2.3"
+                    }
+                    JSON
+                    ,
```
weierophinney:
```suggestion
                    JSON,
```

PHP 7.3 allows following the marker with other symbols, too. :smile:

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -0,0 +1,168 @@
...
```
weierophinney:
```suggestion
                    JSON,
```

test/unit/Github/Api/V3/CreatePullRequestTest.php
```diff
@@ -83,15 +82,15 @@ public function testSuccessfulRequest(): void
...
-JSON
+                    {
+                        "title": "the-title",
+                        "head": "the/source-branch",
+                        "base": "the/target-branch",
+                        "body": "the-body",
+                        "maintainer_can_modify": true,
+                        "draft": false
+                    }
+                    JSON
```
weierophinney:
```suggestion
                    JSON,
```

weierophinney: You'll also need to remove the next line.

test/unit/Github/Api/V3/CreateReleaseTest.php
```diff
@@ -81,12 +80,12 @@ public function testSuccessfulRequest(): void
...
-    "body": "the-body"
-}
-JSON
+                    {
+                        "tag_name": "1.2.3",
+                        "name": "1.2.3",
+                        "body": "the-body"
+                    }
+                    JSON
                     ,
```
weierophinney: Combine these two lines.

weierophinney (changes requested):
Minor tweaks only, but looks good!

geerteltink:
@Ocramius 🚢

weierophinney (approved):
Reads well, I can understand what is happening! Nice work, @geerteltink !

PR #40, opened 2020-08-11
When a new release is made, bump the `CHANGELOG` version in the next release branch and the originating release branch

### LINKED ISSUE #33: Whenever a `CHANGELOG.md` release date is set, also create a new "next release" block with the `TBD` in the date
### Feature Request

Requires #18 

Discussed with @weierophinney: we probably want to prepare the next `CHANGELOG.md` block each time we do a release, so maintainers don't have to manually add the `CHANGELOG.md` template:

```md
## $x.$y+1.0 - TBD

- Nothing.

### Added

- Even moar stuff by @ocramius, but nobody really cares :D 

### Changed

- Nothing.

### Deprecated

- Nothing.

### Removed

- Nothing.

### Fixed

- Nothing.
```

Not sure if we should generate the block for `$x.$y+1.0`, or also for `$x.$y.$z+1` :thinking: 

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

weierophinney:
| Q | A |
| - | - |
| New Feature | yes |
| RFC | #33 |
| BC Break | ? |

This patch adds a new interface, `BumpAndCommitChangelogVersion`, which describes classes that can update a changelog file to add a new version, and then commit and push the changes. I have included one implementation, `BumpAndCommitChangelogVersionViaKeepAChangelog` which uses utilities from phly/keep-a-changelog, as well as internal utilities, to do this.

I have added `BumpAndCommitChangelogVersion` as a dependency on `SwitchDefaultBranchToNextMinor`. The functionality is invoked only if a new release branch is made, and then immediately following that operation. At that time, the changelog is updated with a stub for the next **minor** version.

I have created a new command, `laminas:releases:bump-changelog`, implemented by `BumpChangelogForReleaseBranch`, which determines the release version and release branch, and uses those to call the composed `BumpAndCommitChangelogVersion`, this time bumping the changelog to the next **patch** version on the **originating** release branch.

Finally, I updated the sample workflow to call `laminas:releases:bump-changelog` as the final action in the workflow.  This approach guarantees that the newly-created `vNEXT` release branch does NOT contain a stub for the next patch release; only the originating branch will.

Fixes #33

### DIFF
--- a/bin/console.php
+++ b/bin/console.php
@@ -8,6 +8,7 @@ namespace Laminas\AutomaticReleases\WebApplication;
 use ErrorException;
 use Http\Discovery\HttpClientDiscovery;
 use Http\Discovery\Psr17FactoryDiscovery;
+use Laminas\AutomaticReleases\Application\Command\BumpChangelogForReleaseBranch;
 use Laminas\AutomaticReleases\Application\Command\CreateMergeUpPullRequest;
 use Laminas\AutomaticReleases\Application\Command\ReleaseCommand;
 use Laminas\AutomaticReleases\Application\Command\SwitchDefaultBranchToNextMinor;
@@ -133,6 +134,13 @@ use const STDERR;
                 $httpClient,
                 $githubToken
             ),
+            $bumpChangelogVersion
+        ),
+        new BumpChangelogForReleaseBranch(
+            $variables,
+            $loadEvent,
+            $fetch,
+            $getCandidateBranches,
             $bumpChangelogVersion
         ),
     ]);
--- a/bin/console.php
+++ b/bin/console.php
@@ -11,8 +11,10 @@ use Http\Discovery\Psr17FactoryDiscovery;
 use Laminas\AutomaticReleases\Application\Command\CreateMergeUpPullRequest;
 use Laminas\AutomaticReleases\Application\Command\ReleaseCommand;
 use Laminas\AutomaticReleases\Application\Command\SwitchDefaultBranchToNextMinor;
+use Laminas\AutomaticReleases\Changelog\BumpAndCommitChangelogVersionViaKeepAChangelog;
 use Laminas\AutomaticReleases\Changelog\CommitReleaseChangelogViaKeepAChangelog;
 use Laminas\AutomaticReleases\Environment\EnvironmentVariables;
+use Laminas\AutomaticReleases\Git\CheckoutBranchViaConsole;
 use Laminas\AutomaticReleases\Git\CommitFileViaConsole;
 use Laminas\AutomaticReleases\Git\CreateTagViaConsole;
 use Laminas\AutomaticReleases\Git\FetchAndSetCurrentUserByReplacingCurrentOriginRemote;
@@ -66,6 +68,7 @@ use const STDERR;
         $httpClient,
         $githubToken
     ));
+    $checkoutBranch       = new CheckoutBranchViaConsole();
     $commit               = new CommitFileViaConsole();
     $push                 = new PushViaConsole();
     $commitChangelog      = new CommitReleaseChangelogViaKeepAChangelog(new SystemClock(), $commit, $push, $logger);
@@ -81,6 +84,12 @@ use const STDERR;
         $makeRequests,
         $httpClient,
         $githubToken
+    );
+    $bumpChangelogVersion = new BumpAndCommitChangelogVersionViaKeepAChangelog(
+        $checkoutBranch,
+        $commit,
+        $push,
+        $logger
     );
 
     /** @psalm-suppress DeprecatedClass */
@@ -123,7 +132,8 @@ use const STDERR;
                 $makeRequests,
                 $httpClient,
                 $githubToken
-            )
+            ),
+            $bumpChangelogVersion
         ),
     ]);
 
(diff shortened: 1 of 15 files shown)


examples/.github/workflows/release-on-milestone-closed.yml
```diff
@@ -45,3 +45,13 @@ jobs:
           "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
           "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Bump Changelog Version On Originating Release Branch"
```
Ocramius, line 49: Should be done also in `.github`, not just in examples

src/Changelog/BumpAndCommitChangelogVersion.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Changelog;
+
+use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
+
+interface BumpAndCommitChangelogVersion
+{
+    public const BUMP_MINOR = 'bumpMinorVersion';
```
Ocramius, line 12: Not a fan of these public constants to switch behaviour

src/Changelog/BumpAndCommitChangelogVersion.php
```diff
@@ -0,0 +1,30 @@
...
+    public const BUMP_PATCH = 'bumpPatchVersion';
+
+    public const KNOWN_BUMP_TYPES = [
+        self::BUMP_MINOR,
+        self::BUMP_PATCH,
+    ];
+
+    /**
+     * @psalm-param value-of<self::KNOWN_BUMP_TYPES> $bumpType
```
Ocramius: Can be `self::BUMP_*` (actual type syntax), so we don't need to declare the array

src/Changelog/BumpAndCommitChangelogVersion.php
Ocramius: Hmm, unsure about this one: possibly good enough, since we enforce it at type level anyway 🤔

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
```diff
@@ -0,0 +1,85 @@
...
+
+    public function __invoke(
+        string $bumpType,
+        string $repositoryDirectory,
+        SemVerVersion $version,
+        BranchName $sourceBranch
+    ): void {
+        ($this->checkoutBranch)($repositoryDirectory, $sourceBranch->name());
+
+        $changelogFile = sprintf('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
```
Ocramius, line 54: Not a problem, but sprintf for concatenating `/`? 😅

src/Git/CheckoutBranch.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Git;
+
+interface CheckoutBranch
+{
+    /**
+     * @psalm-param non-empty-string $repositoryDirectory
+     * @psalm-param non-empty-string $branchName
```
Ocramius: We have a value type for `BranchName` in the repo: let's use that 👍

src/Git/CheckoutBranchViaConsole.php
```diff
@@ -0,0 +1,18 @@
...
+use Symfony\Component\Process\Process;
+
+class CheckoutBranchViaConsole implements CheckoutBranch
+{
+    public function __invoke(
+        string $repositoryDirectory,
+        string $branchName
+    ): void {
+        (new Process(['git', 'checkout', $branchName], $repositoryDirectory))
+            ->mustRun();
```
Ocramius, line 17: I wonder if we should use `switch` these days

bin/console.php
```diff
@@ -123,7 +133,15 @@ static function (int $errorCode, string $message = '', string $file = '', int $l
                 $makeRequests,
                 $httpClient,
                 $githubToken
-            )
+            ),
+            $bumpChangelogVersion
+        ),
+        new BumpChangelogForReleaseBranch(
```
Ocramius, line 139: I like that this is a new command to be explicitly run after the release 👍

Ocramius (changes requested):

examples/.github/workflows/release-on-milestone-closed.yml
weierophinney: oops!

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
weierophinney: I find it easier to read. :stuck_out_tongue:

Ocramius: That's alright, just exotic 👍

src/Git/CheckoutBranch.php
weierophinney: I'd actually originally done that, but reverted because another class was taking just the string value. Updating again...

src/Git/CheckoutBranchViaConsole.php
weierophinney: Yes, we should. I'll update here, and elsewhere.

weierophinney: Actually... will only update in the stuff relevant to this particular patch. I'll do a separate PR to update this everywhere later.

weierophinney:
@Ocramius I've pushed changes addressing the feedback!

Ocramius (approved):
Excellent, thanks @weierophinney!

PR #41, opened 2020-08-12
Do not checkout the branch passed to `CommitFileViaConsole`

weierophinney:
We need to assume we're on the correct branch; otherwise, a checkout operation will lead to an error, aborting the operation.

Instead, I've added functionality to verify that we are on the branch provided; if not, an exception is raised.

The changes also required annotating that the repository directory, filename, and commit message arguments could not be empty, which required changes elsewhere to ensure this is true before calling on `CommitFile`.

### DIFF
--- a/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
+++ b/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
@@ -67,7 +67,7 @@ class BumpAndCommitChangelogVersionViaKeepAChangelog implements BumpAndCommitCha
         $bumper->updateChangelog($newVersion);
 
         $message = sprintf(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion);
-        Assert::stringNotEmpty($message);
+        Assert::notEmpty($message);
 
         ($this->commitFile)(
             $repositoryDirectory,
--- a/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
+++ b/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
@@ -66,11 +66,14 @@ class BumpAndCommitChangelogVersionViaKeepAChangelog implements BumpAndCommitCha
         Assert::stringNotEmpty($newVersion);
         $bumper->updateChangelog($newVersion);
 
+        $message = sprintf(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion);
+        Assert::stringNotEmpty($message);
+
         ($this->commitFile)(
             $repositoryDirectory,
             $sourceBranch,
             self::CHANGELOG_FILE,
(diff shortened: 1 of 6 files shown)


src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
```diff
@@ -66,11 +66,14 @@ public function __invoke(
         Assert::stringNotEmpty($newVersion);
         $bumper->updateChangelog($newVersion);
 
+        $message = sprintf(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion);
+        Assert::stringNotEmpty($message);
```
Ocramius: I think this is just `Assert::notEmpty()`, since `sprintf(): string` already enforces it to be a `string` 

Doesn't psalm complain about this one?

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
```diff
@@ -69,11 +70,14 @@ public function __invoke(
             return;
         }
 
+        $message = sprintf(self::COMMIT_TEMPLATE, $versionString, self::CHANGELOG_FILE);
+        Assert::stringNotEmpty($message);
```
Ocramius: Same comment as above: need to `Assert::notEmpty()` only

src/Git/CommitFileViaConsole.php
```diff
@@ -26,4 +29,30 @@ public function __invoke(
             $repositoryDirectory
         ))->mustRun();
     }
+
+    /**
+     * @param non-empty-string $repositoryDirectory
+     * @param non-empty-string $filename
+     */
+    private function assertBranch(
```
Ocramius: `assertBranch()` naming is a bit generic: would suggest naming this something like `assertWeAreOnBranch...` or similar

src/Git/CommitFileViaConsole.php
```diff
@@ -26,4 +29,30 @@ public function __invoke(
...
+        BranchName $expectedBranch,
+        string $repositoryDirectory,
+        string $filename
+    ): void {
+        $process = new Process(['git', 'branch', '--show-current'], $repositoryDirectory);
+        $process->run();
```
Ocramius: You want to call `$process->mustRun()` here: that already throws meaningful exceptions with STDOUT/STDERR as part of them (no need for the assertion either)

Ocramius (changes requested):

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
weierophinney: It didn't - it complained that it couldn't determine it was non-empty, which is why I added the assertion. I'll update to simply `notEmpty()` and see what happens.

weierophinney:
Feedback incorporated!

Ocramius (approved):
LGTM :+1:

PR #42, opened 2020-08-12
Update Dockerfile to have modern toolchain

weierophinney:
We need a more recent git version, with 1.22 as the minimum. This patch updates the Dockerfile to use Ubuntu 20.04 as base image, adds the ondrej repository, and installs PHP with the extensions we need in order to run the tooling.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM composer:1 AS composer
 
-FROM php:7.4-cli
+FROM ubuntu:20.04
 
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
 WORKDIR /app
 
 RUN apt update \
+    && apt install -y software-properties-common \
+    && apt clean
+
+RUN add-apt-repository -y ppa:ondrej/php \
     && apt install -y \
         git \
         gnupg \
         libzip-dev \
         zip \
-    && docker-php-ext-install zip \
+        php7.4-cli \
+        php7.4-curl \
+        php7.4-json \
+        php7.4-mbstring \
+        php7.4-readline \
+        php7.4-xml \
+        php7.4-zip \
     && apt clean
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,7 @@ WORKDIR /app
 
 RUN apt update \
     && apt install -y software-properties-common \
-    && apt clean
-
-RUN add-apt-repository -y ppa:ondrej/php \
+    && add-apt-repository -y ppa:ondrej/php \
     && apt install -y \
         git \
         gnupg \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
     && apt install -y \
         git \
         gnupg \
         libzip-dev \
         zip \
-    && docker-php-ext-install zip \
+        php7.4-cli \
+        php7.4-curl \
+        php7.4-json \
+        php7.4-mbstring \
```
weierophinney, line 29: This is required by the jwage changelog generator.

Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
+RUN add-apt-repository -y ppa:ondrej/php \
     && apt install -y \
         git \
         gnupg \
         libzip-dev \
         zip \
-    && docker-php-ext-install zip \
+        php7.4-cli \
+        php7.4-curl \
+        php7.4-json \
```
weierophinney, line 28: We're parsing JSON. I really wish this were just on by default.

Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
+
+RUN add-apt-repository -y ppa:ondrej/php \
     && apt install -y \
         git \
         gnupg \
         libzip-dev \
         zip \
-    && docker-php-ext-install zip \
+        php7.4-cli \
+        php7.4-curl \
```
weierophinney, line 27: IIRC, your client uses cURL.

Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
+        php7.4-json \
+        php7.4-mbstring \
+        php7.4-readline \
```
weierophinney, line 30: Needed by composer and symfony/console.

Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
+        php7.4-xml \
```
weierophinney, line 31: Needed by phly/keep-a-changelog

Dockerfile
```diff
@@ -16,12 +16,22 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
 WORKDIR /app
 
 RUN apt update \
+    && apt install -y software-properties-common \
```
weierophinney, line 19: There might be a way to install the ondrej PPA without using `add-apt-repository`; this was just the easiest way I could figure out.

Ocramius: Yeah, seems to be the standard way.

Can you please though put it in a single layer with the next `RUN`? Otherwise, we get two layers moving around a lot of `apt` stuff, potentially unnecessarily

Ocramius (approved):
:ship:

PR #45, opened 2020-08-12
Fix YAML config for actions-tagger workflow

weierophinney:
Updated how the "types" configuration for the release action are defined, and limited to only "published".

Additionally, working examples I saw all used a different format for defining env variables, so I went with what appears to work.


weierophinney:
Going to re-target 1.3.x, as github uses the default branch to figure out which workflows to run.

Issue #46, opened 2020-08-12
Does not checkout release branch before attempting to update changelog


weierophinney:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 1.2.1

#### Summary

The `laminas:automatic-releases:release` command fails when attempting to commit the release date to the `CHANGELOG.md` file when the default branch is other than the release branch.

As a specific example: I recently attempted to close milestone 2.12.3 of laminas-mail. A 2.13.x branch already exists. When it attempted to push changes for the changelog, I received the error here: https://github.com/laminas/laminas-mail/runs/977187215?check_suite_focus=true#step:4:14

#### How to reproduce

Create a release branch "X.Y.x", and another "X.Y+1.x". Close a milestone related to "X.Y.x".

#### Expected behavior

It should check out the associated release branch prior to updating the changelog so that it can commit and push to it.

Ocramius:
Handled in #47

PR #47, opened 2020-08-12
Idempotent operations for reading, checkout branch prior to writing

weierophinney:
| Q | A |
| - | - |
| BC Break? | No |
| Bug?      | Yes — #46 |

This patch does the following:

- Adds  the interface `ChangelogExists`, with one implementation, `ChangelogExistsViaConsole`. The functionality is intended to be used whenever we want to check for existence of a changelog on a given branch, and is idempotent (it uses the exit code of `git show {REF}:CHANGELOG.md` to determine existence).

- Any operation that needs to check for existence of the changelog now composes an instance of `ChangelogExists` for that purpose.

- Operations that only _read_  the `CHANGELOG.md` now do so using `git show {REF}:CHANGELOG.md`.This change was introduced to  `CreateReleaseTextViaKeepAChangelog`.

- Any operation that will _write_ to the CHANGELOG.md now first uses `ChangelogExists` to test for the file, and then performs a `CheckoutBranch` operation to ensure that the correct version is updated, and to ensure we push back to the correct branch. Specifically, I updated `CreateReleaseTextViaKeepAChangelog` to compose a `CheckoutBranch` instance for use after determining changes can be made.

This should resolve all remaining issues regarding updates to the changelog file.

### DIFF
--- a/src/Changelog/ChangelogExistsViaConsole.php
+++ b/src/Changelog/ChangelogExistsViaConsole.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Changelog;
+
+use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Symfony\Component\Process\Process;
+
+class ChangelogExistsViaConsole implements ChangelogExists
+{
+    /**
+     * @param non-empty-string $repositoryDirectory
+     */
+    public function __invoke(
+        BranchName $sourceBranch,
+        string $repositoryDirectory
+    ): bool {
+        $process = new Process(['git', 'show', $sourceBranch->name() . ':CHANGELOG.md'], $repositoryDirectory);
+        $process->run();
+
+        return $process->isSuccessful();
+    }
+}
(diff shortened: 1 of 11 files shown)


weierophinney:
@Ocramius all yours to review. :)

src/Github/CreateReleaseTextViaKeepAChangelog.php
```diff
@@ -5,19 +5,25 @@
...
 use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
 use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetMilestoneChangelog\Response\Milestone;
 use Laminas\AutomaticReleases\Github\Value\RepositoryName;
 use Phly\KeepAChangelog\Common\ChangelogParser;
 use Phly\KeepAChangelog\Exception\ExceptionInterface;
+use Symfony\Component\Process\Process;
 use Webmozart\Assert\Assert;
 
-use function file_exists;
-use function file_get_contents;
```
Ocramius, line 17: I like that these filesystem access operations went away from here 👍

Ocramius (approved):
Matches discussions, improves state management too 👍

:shipit:

src/Changelog/ChangelogExistsViaConsole.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * @param non-empty-string $repositoryDirectory
+     */
+    public function __invoke(
+        BranchName $sourceBranch,
+        string $repositoryDirectory
+    ): bool {
+        $process = new Process(['git', 'show', $sourceBranch->name() . ':CHANGELOG.md'], $repositoryDirectory);
+        $process->run();
```
weierophinney, line 20: Very specifically not "mustRun()" here, as failure is an expected status.

PR #50, opened 2020-08-13
When using the REST API, send "token" authorization instead of "bearer"

weierophinney:
Since our GITHUB_TOKEN is a personal access token, it can be used as either an OAuth (bearer) or PAT (token) type for purposes of authorization. However, OAuth requests from actions do not trigger workflow events (see https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token).

This patch updates the three different API calls we currently perform (create release, set default branch, create pull request) to use the "token" type for authorization.


Ocramius (approved):

Ocramius:
Let's try it! 🚢

PR #54, opened 2020-08-13
Update workflow documentation and examples

weierophinney:
- Use a standard `GITHUB_TOKEN` for the `Bump Changelog Version ..." step.
- Document why you might want to use a PAT with the "Release" step, and how to accomplish that.

**NO MILESTONE, NO RELEASE NEEDED**; this is just documentation.


README.md
```diff
@@ -35,6 +35,26 @@ to your project or organization:
   command, because [changing default branch of a repository currently requires administrative token rights](https://developer.github.com/v3/repos/#update-a-repository).
   You can generate a token from your [personal access tokens page](https://github.com/settings/tokens/new).
 
+> ### Triggering release workflow events
```
Ocramius: Is the quoting intentional?

README.md
```diff
@@ -35,6 +35,26 @@ to your project or organization:
...
+> ```yaml
+>     - name: "Release"
+>       uses: "./"
+>       with:
+>         command-name: "laminas:automatic-releases:release"
+>       env:
+>         "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+> ```
+
 ## Usage
```
Ocramius: I think `Usage` should come before the `Triggering release workflow events`, which is a bit of a niche

README.md
weierophinney: Wanted it to stand out. I can remove if you want.

README.md
weierophinney: Reasonable - will push in a moment.

README.md
Ocramius: Really looks like an edge case - semantically, quoting means something very different than "standing out" :D

Ocramius (approved):

PR #58, opened 2020-08-25
Automatically generate changelog revision text if the changelog was not hand-crafted - do not commit if unchanged

weierophinney:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | code, yes; UI and usage, no
| New Feature   | yes and no
| RFC           | yes

### Description

When rolling out automatic-releases to various repositories, we noticed an issue with the workflow for documentation, mainly that our docs-build action only builds on pushes to the "master" branch, we are trying to eliminate. This has meant manually pushing the current release branch to the "master" branch on the repository whenever there are changes.

@Ocramius suggested that we merge documentation changes just like bugfixes: assign them to milestones, and build on release. The objections raised were:

- We would have unchanged packages from one version to the next, as docs are excluded via `.gitattributes`. We decided this was not really an issue; most people are not updating regularly, and a package that introduces zero code changes is generally welcome as it represents no risk.

- We would have empty changelogs and release announcements.

This latter was something I realized we could address in the automatic-releases tooling, as:

- We already generate additional release notes via jwage/changelog-generator.
- The phly/keep-a-changelog tooling does not require that entries are in a specific format unless you want to use its `entry:*` commands; the only requirements are the version headings.

As such, this patch refactors how changelogs are generated for release notes, and then committed back to the repository.

In particular:

- It introduces `Laminas\AutomaticReleases\Changelog\ChangelogReleaseNotes`, which is an entity that composes the changelog contents, and optionally a phly/keep-a-changelog `ChangelogEntry` instance. It has behavior:
  - `requiresUpdatingChangelogFile()` returns true if a `ChangelogEntry` is composed AND its contents differ from those stored in the `ChangelogReleaseNotes` instance.
  - `writeChangelogFile(string $changelogFile)` will write the release notes contents to the specified changelog file, per the metadata in the `ChangelogEntry` instance composed.
  - `merge(ChangelogReleaseNotes $next): ChangelogReleaseNotes` will clone itself and merge `$next` into it, returning the clone.

- I updated `CreateReleaseText` to have it return a `ChangelogReleaseNotes` instance.

- I updated `CreateReleaseTextViaKeepAChangelog` to now mimics how the phly/keep-a-changelog library parses a changelog file when parsing string contents representing a full changelog as pulled from the originating branch, in order to create a `ChangelogEntry` that details where the original is found in the file, and how much space it takes. Additionally, it then sets the date for the version, and checks to see if the contents have changed from the template; if not, it clears them before passing them to the `ChangelogReleaseNotes` instance it creates.

- I updated `CreateReleaseThroughChangelog` to perform some minor markdown formatting changes to better accommodate being injected in a CHANGELOG.md file; it also now returns a `ChangelogReleaseNotes` instance (without a `ChangelogEntry` composed).

- I updated `CommitReleaseChangelog` to accept a `ChangelogReleaseNotes` as the first argument; implementations now use it to actually write the `CHANGELOG.md` file prior to committing and pushing it.

- I renamed `ConcatenateMultipleReleaseTexts` to `MergeMultipleReleaseNotes` to better describe what it now does (it uses the `merge()` operation of returned `ChangelogReleaseNotes` instances, returning the final product).

- I updated the logic of `ReleaseCommand` to generate release text first, then commit it.

## tl;dr

When the "release" command of automatic-releases now runs, it will:

- Always generate release notes using jwage/changelog-generator.
- **IF** a `CHANGELOG.md` file is present, it then:
  - Checks to see if it contains any changes for the target version. If not, it replaces the content with the generated release notes.
  - Otherwise, it appends the content with the generated release notes.

The upshot is that documentation maintainers can:

- Add documentation patches to the appropriate milestone.
- Merge those changes.
- Close the milestone.

and get a release issued, which will then trigger builds for documentation.

## Final note

We will need to use the `ORGANIZATION_ADMIN_TOKEN` for the value of the `GITHUB_TOKEN` on the `Release` step of the workflow in order to trigger the release workflow event, and thus the documentation build.

### DIFF
--- a/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
+++ b/examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
@@ -1,0 +1,57 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Automatic Releases"
+
+on:
+  milestone:
+    types:
+      - "closed"
+
+jobs:
+  release:
+    name: "GIT tag, release & create merge-up PR"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Release"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:release"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create Merge-Up Pull Request"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:create-merge-up-pull-request"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Create and/or Switch to new Release Branch"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.ORGANIZATION_ADMIN_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
+
+      - name: "Bump Changelog Version On Originating Release Branch"
+        uses: "laminas/automatic-releases@v1"
+        with:
+          command-name: "laminas:automatic-releases:bump-changelog"
+        env:
+          "GITHUB_TOKEN": ${{ secrets.GITHUB_TOKEN }}
+          "SIGNING_SECRET_KEY": ${{ secrets.SIGNING_SECRET_KEY }}
+          "GIT_AUTHOR_NAME": ${{ secrets.GIT_AUTHOR_NAME }}
+          "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
--- a/src/Application/Command/ReleaseCommand.php
+++ b/src/Application/Command/ReleaseCommand.php
@@ -84,9 +84,7 @@ final class ReleaseCommand extends Command
             sprintf('No valid release branch found for version %s', $releaseVersion->fullReleaseName())
         );
 
-        ($this->commitChangelog)($repositoryPath, $releaseVersion, $releaseBranch);
-
-        $changelog = ($this->createChangelogText)(
+        $changelogReleaseNotes = ($this->createChangelogText)(
             $milestone,
             $repositoryName,
             $releaseVersion,
@@ -94,12 +92,20 @@ final class ReleaseCommand extends Command
             $repositoryPath
         );
 
+        ($this->commitChangelog)($changelogReleaseNotes, $repositoryPath, $releaseVersion, $releaseBranch);
+
         $tagName = $releaseVersion->fullReleaseName();
 
-        ($this->createTag)($repositoryPath, $releaseBranch, $tagName, $changelog, $this->environment->signingSecretKey());
+        ($this->createTag)(
+            $repositoryPath,
+            $releaseBranch,
+            $tagName,
+            $changelogReleaseNotes->contents(),
+            $this->environment->signingSecretKey()
+        );
         ($this->push)($repositoryPath, $tagName);
         ($this->push)($repositoryPath, $tagName, $releaseVersion->targetReleaseBranchName()->name());
-        ($this->createRelease)($repositoryName, $releaseVersion, $changelog);
+        ($this->createRelease)($repositoryName, $releaseVersion, $changelogReleaseNotes->contents());
 
         return 0;
     }
--- a/src/Changelog/ChangelogReleaseNotes.php
+++ b/src/Changelog/ChangelogReleaseNotes.php
@@ -9,6 +9,9 @@ use Phly\KeepAChangelog\Common\ChangelogEntry;
 use RuntimeException;
 use Webmozart\Assert\Assert;
 
+/**
+ * @psalm-immutable
+ */
 class ChangelogReleaseNotes
 {
     private const CONCATENATION_STRING = "\n\n-----\n\n";
@@ -19,13 +22,39 @@ class ChangelogReleaseNotes
     private string $contents;
 
     /**
+     * @psalm-param non-empty-string $changelogFile
+     */
+    public static function writeChangelogFile(string $changelogFile, self $releaseNotes): void
+    {
+        // Nothing to do
+        if (! $releaseNotes->requiresUpdatingChangelogFile()) {
+            return;
+        }
+
+        Assert::notNull($releaseNotes->changelogEntry);
+
+        $editor = new ChangelogEditor();
+        $editor->update(
+            $changelogFile,
+            $releaseNotes->contents,
+            $releaseNotes->changelogEntry
+        );
+    }
+
+    /**
      * @psalm-param non-empty-string $contents
      */
     public function __construct(
         string $contents,
         ?ChangelogEntry $changelogEntry = null
     ) {
-        $this->contents       = $contents;
+        if ($changelogEntry) {
+            $changelogEntry = clone $changelogEntry;
+        }
+
+        $this->contents = $contents;
+
+        /** @psalm-suppress ImpurePropertyAssignment */
         $this->changelogEntry = $changelogEntry;
     }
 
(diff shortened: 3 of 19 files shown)


src/Changelog/ChangelogReleaseNotes.php
```diff
@@ -0,0 +1,86 @@
...
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Changelog;
+
+use Phly\KeepAChangelog\Common\ChangelogEditor;
+use Phly\KeepAChangelog\Common\ChangelogEntry;
+use RuntimeException;
+use Webmozart\Assert\Assert;
+
+class ChangelogReleaseNotes
```
Ocramius, line 15: I'd love for this to be `@psalm-immutable`, but we do perform filesystem interactions below.

It is still possible to achieve that by making the method that performs side effects `static`, but that isn't the end of the world, so I'll leave it up to the author's discretion

src/Application/Command/ReleaseCommand.php
```diff
@@ -84,22 +84,28 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
-        $changelog = ($this->createChangelogText)(
+        $changelogReleaseNotes = ($this->createChangelogText)(
             $milestone,
             $repositoryName,
             $releaseVersion,
             $releaseBranch,
             $repositoryPath
         );
 
+        ($this->commitChangelog)($changelogReleaseNotes, $repositoryPath, $releaseVersion, $releaseBranch);
```
Ocramius, line 95: I like how this now receives a pre-computed changelog 👍

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
```diff
@@ -90,104 +86,4 @@ public function __invoke(
 
         ($this->push)($repositoryDirectory, $sourceBranch->name());
     }
-
-    private function updateChangelog(string $changelogFile, string $versionString): bool
```
Ocramius, line 94: Excellent refactoring 💪

src/Github/CreateReleaseTextThroughChangelog.php
```diff
@@ -73,4 +80,36 @@ private function markdownLink(string $text, UriInterface $uri): string
     {
         return '[' . $text . '](' . $uri->__toString() . ')';
     }
+
+    private function normalizeChangelogHeadings(string $changelog): string
+    {
+        $lines         = explode("\n", trim($changelog));
+        $lineCount     = count($lines);
+        $linesToRemove = [];
+        for ($i = 0; $i < $lineCount; $i += 1) {
```
Ocramius: This code block is really hard to understand. I suppose most of it is tested well, according to mutation testing results, but it is really making the reader's head spin.

Refactoring can be deferred, since the code is functionally pure, but we need comments about what is being done here.

src/Github/CreateReleaseTextThroughChangelog.php
```diff
@@ -73,4 +80,36 @@ private function markdownLink(string $text, UriInterface $uri): string
...
+            $line    = $lines[$i];
+            $matches = [];
+            if (! preg_match('/^(?P<delim>-{3,}|={3,})$/', $line, $matches)) {
+                continue;
+            }
+
+            $previousLine = $lines[$i - 1];
+            if (empty($previousLine)) {
```
Ocramius, line 117: What happens with `$i = 0`? I expect a notice in the line above 🤔

src/Github/CreateReleaseTextViaKeepAChangelog.php
```diff
@@ -84,4 +128,78 @@ private function fetchChangelogContentsFromBranch(
...
+     */
+    private function fetchChangelogEntry(string $contents, string $version): ChangelogEntry
+    {
+        $entryContents = [];
+        $entryIndex    = null;
+        $entryLength   = 0;
+        $boundaryRegex = '/^(?:## (?:\d+\.\d+\.\d+|\[\d+\.\d+\.\d+\])|\[.*?\]:\s*\S+)/i';
+        $regex         = sprintf('/^## (?:%1$s|\[%1$s\])/i', preg_quote($version));
+
+        foreach (explode("\n", $contents) as $index => $line) {
```
Ocramius, line 210: Similar to other changelog normalisation logic, this is really hard to follow.

Can defer refactoring, but need comments about what is being done in this method.

src/Github/MergeMultipleReleaseNotes.php
```diff
@@ -33,19 +32,24 @@ public function __invoke(
...
 
-        $changelog = implode(self::CONCATENATION_STRING, $items);
-        Assert::notEmpty($changelog);
+        $releaseNotes = array_reduce(
+            $items,
+            static fn (?ChangelogReleaseNotes $releaseNotes, ChangelogReleaseNotes $item): ChangelogReleaseNotes => $releaseNotes ? $releaseNotes->merge($item) : $item
+        );
+
+        Assert::isInstanceOf($releaseNotes, ChangelogReleaseNotes::class);
+        Assert::notEmpty($releaseNotes->contents());
```
Ocramius: Is this assertion needed? I think `->contents()` already returns a `non-empty-string`

test/unit/Changelog/ChangelogReleaseNotesTest.php
```diff
@@ -0,0 +1,222 @@
...
+use function array_slice;
+use function explode;
+use function file_get_contents;
+use function file_put_contents;
+use function implode;
+use function Safe\tempnam;
+use function sprintf;
+use function sys_get_temp_dir;
+
+class ChangelogReleaseNotesTest extends TestCase
```
Ocramius, line 26: Consider adding `@covers`: makes mutation testing more precise on the SUT

test/unit/Github/CreateChangelogTextTest.php
```diff
@@ -10,12 +10,14 @@
...
 use Laminas\AutomaticReleases\Github\GenerateChangelog;
 use Laminas\AutomaticReleases\Github\Value\RepositoryName;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 
 final class CreateChangelogTextTest extends TestCase
 {
     public function testGeneratedReleaseText(): void
     {
+        /** @psalm-var GenerateChangelog&MockObject $generateChangelog */
```
Ocramius: This inline type shouldn't be needed: phpunit mocking utilities already have the correct annotations

Ocramius (changes requested):
One potential bug, some space for refactoring, otherwise requires commenting on tricky logic that parses release notes.

I think this needs only quick adjustments, and then we can release it as `1.3.0`

src/Github/CreateReleaseTextThroughChangelog.php
samsonasik: how about using `foreach` instead ? 

```suggestion
        foreach ($lines as $i => $line) {
```

We can eliminate usage of `$lineCount     = count($lines);`

src/Github/CreateReleaseTextThroughChangelog.php
```diff
@@ -73,4 +80,36 @@ private function markdownLink(string $text, UriInterface $uri): string
     {
         return '[' . $text . '](' . $uri->__toString() . ')';
     }
+
+    private function normalizeChangelogHeadings(string $changelog): string
+    {
+        $lines         = explode("\n", trim($changelog));
+        $lineCount     = count($lines);
```
samsonasik: can be removed if next `for` changed to `foreach`

src/Github/CreateReleaseText.php
```diff
@@ -13,15 +14,14 @@ interface CreateReleaseText
...
-     * @psalm-return non-empty-string
      */
     public function __invoke(
         Milestone $milestone,
         RepositoryName $repositoryName,
         SemVerVersion $semVerVersion,
         BranchName $sourceBranch,
         string $repositoryDirectory
-    ): string;
+    ): ChangelogReleaseNotes;
```
michalbundyra, line 24: BC Break 😱 Technically then this feature should go to v2.0

michalbundyra (changes requested):

src/Github/CreateReleaseText.php
Ocramius: This package has `"type": "application"` and is not published to packagist either - we can break internals as much as we want (and feel free to do so!)

src/Application/Command/ReleaseCommand.php
weierophinney: I thought you might. :smile:

test/unit/Github/CreateChangelogTextTest.php
weierophinney: This was in part for intelephense code completion.

src/Application/Command/ReleaseCommand.php
weierophinney: This made me happy as well. The "commit, then create" from before worked, and was necessary based on what we were doing, but this makes sooooo much more sense.

src/Changelog/ChangelogReleaseNotes.php
weierophinney: So, I tried doing this just now, and the problem I ran into is that `ChangelogEntry` is mutable, and Psalm doesn't like having an externally mutable instance inside an immutable object. I tried cloning it, but it somehow fails to see that there's a clone created internally, and that the instance is not exposed publicly at any point (there's no getter/setter for it).

I think I'll move forward with it, but will need to suppress the `ImpurePropertyAssignment` warnings, knowing that they were being raised incorrectly. (Just letting you know in advance.)

test/unit/Github/CreateChangelogTextTest.php
Ocramius: Doesn't intelephense read something like `.phpstorm.meta.php`?

src/Changelog/ChangelogReleaseNotes.php
Ocramius: Ah, I see, `ChangelogEntry` is indeed external. Missed that, sorry!

Overall, I'm OK with suppressing those for now, so this version of the code should be fine.

src/Github/CreateReleaseTextThroughChangelog.php
Ocramius: Fixed by conditional introduced above :+1:

Ocramius (approved):
LGTM :ship:

Ocramius:
Ah, can't merge this due to mutation score going down quite a lot. This is most likely due to the complex logic in the changelog text generation/extraction code: we can either test it more carefully (more string examples, effectively), or rewrite bits of it to be more functional-oriented, and removing prodcedural/iteration bits that are more mutation-prone (yes, FP has the advantage of reducing code paths :D )

weierophinney:
Mutation score fixed, @Ocramius :smile:

examples/.github/workflows/release-on-milestone-closed-triggering-release-event.yml
```diff
@@ -0,0 +1,57 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
```
Ocramius, line 1: This example needs a better comment

Ocramius (approved):
LGTM - the documentation of the new example can wait a bit

Issue #64, opened 2020-09-03
Normalize generated changelog to remove redundant information and whitespace


weierophinney:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | enhancement
| RFC         | no
| BC Break    | no

#### Summary

Currently, the jwage/changelog-generator -generated content includes:

```markdown

-----

### Release Notes for [1.0.0](https://github.com/mezzio/mezzio-cors/milestone/1)



### 1.0.0
```

Note:

- the extra line breaks
- the repeated "1.0.0" references

The suggestion is to normalize this to:

- never have more than one empty line at a time
- remove the extraneous version references

PR #68, opened 2020-09-03
Strip empty sections from keep-a-changelog release notes

### LINKED ISSUE #63: Remove empty changelog sections when creating changelog from Keep A Changelog format
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | improvement
| RFC         | no
| BC Break    | no

#### Summary

Currently when using Keep A Changelog-formatted changelogs, it's not uncommon to have entries that look like this:

```markdown
## X.Y.Z - TBD

### Added

- Nothing.

### Changed

- Nothing.

### Removed

- Nothing.

### Deprecated

- Nothing.

### Fixed

- Something we fixed
```

This leads to a lot of noise in the generated changelogs, as we have a lot of "Nothing" entries, followed by a list of things we actually did.

The suggested improvement is to find `### {heading}\n\nNothing.\n` entries and remove them as part of changelog fetching and normalization.

weierophinney:
Instead of only looking for and removing "default contents" (i.e., what is generated via keep-a-changelog `bump*` operations), this patch looks for any of the default sections that have the `- Nothing` default contents and removes them.

Fixes #63

### DIFF
--- a/src/Github/CreateReleaseTextViaKeepAChangelog.php
+++ b/src/Github/CreateReleaseTextViaKeepAChangelog.php
@@ -18,6 +18,7 @@ use Phly\KeepAChangelog\Exception\ExceptionInterface;
 use Symfony\Component\Process\Process;
 use Webmozart\Assert\Assert;
 
+use function array_reduce;
 use function count;
 use function explode;
 use function implode;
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
LGTM, but need @michalbundyra to skim over this first

weierophinney:
@michalbundyra There will still be some redundancy between stuff in the main changelog and the generated contents, but the former will give user-focused context that the latter does not. What this patch does, however, is remove the verbosity of that "Nothing" boilerplate, which should make this look better.

The fixes for #64 address the rest of it, which is redundancy in headings and whatnot introduced by the release notes generator. The combination of the the two should give us something you're more comfortable with.

michalbundyra (approved):
🚢

PR #72, opened 2020-09-05
Sign commits, as well as tags

ramsey:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Currently, automatic-releases signs Git tags using the signing key but not commits. This PR provides support for signing commits, as well.

This is a BC break because some of the interfaces have changed to require `Laminas\AutomaticReleases\Gpg\SecretKeyId`. This is not an optional parameter. As such, this change might need to wait until 2.0.0, unless the risk is deemed acceptable for 1.7.0 or 1.8.0.

### DIFF
--- a/test/unit/Git/CommitFileViaConsoleTest.php
+++ b/test/unit/Git/CommitFileViaConsoleTest.php
@@ -7,10 +7,13 @@ namespace Laminas\AutomaticReleases\Test\Unit\Git;
 use InvalidArgumentException;
 use Laminas\AutomaticReleases\Git\CommitFileViaConsole;
 use Laminas\AutomaticReleases\Git\Value\BranchName;
+use Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromStringViaTemporaryFile;
+use Laminas\AutomaticReleases\Gpg\SecretKeyId;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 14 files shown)


geerteltink (approved):

test/unit/Git/CommitFileViaConsoleTest.php
```diff
@@ -93,6 +106,12 @@ public function testFailsIfNotOnCorrectBranch(): void
         $this->expectException(InvalidArgumentException::class);
         $this->expectExceptionMessage('different branch');
         (new CommitFileViaConsole())
-            ->__invoke($this->checkout, BranchName::fromName('1.0.x'), 'README.md', 'commit message');
+            ->__invoke(
```
Ocramius, line 110: I think we meed a test that verifies that the commit is signed

Ocramius (changes requested):

test/unit/Git/CommitFileViaConsoleTest.php
ramsey: Assertion added!

Ocramius:
FWIW, not a BC break, since this is an application, not a library 👍

PR #75, opened 2020-09-12
Fix docker actions name label

geerteltink:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | maybe

### Description

This changes the docker github actions name label. I'm not sure if this can be renamed like this or this has a greater impact in case gihub uses it somewhere.


weierophinney:
GitHub only uses the metadata file, not the Docker labels, in order to differentiate actions, so this change will have no effect on publication. It just makes it more correct. :+1:

PR #76, opened 2020-09-12
Add more logging

geerteltink:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

This PR ads more logging.

### DIFF
--- a/test/unit/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelogTest.php
+++ b/test/unit/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelogTest.php
@@ -49,7 +49,7 @@ class BumpAndCommitChangelogVersionViaKeepAChangelogTest extends TestCase
         $this->push                   = $this->createMock(Push::class);
         $this->logger                 = $this->createMock(LoggerInterface::class);
         $this->bumpAndCommitChangelog = new BumpAndCommitChangelogVersionViaKeepAChangelog(
-            new ChangelogExistsViaConsole(),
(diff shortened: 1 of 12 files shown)


test/unit/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelogTest.php
```diff
@@ -225,29 +228,29 @@ private function createMockChangelog(): string
 
     private const CHANGELOG_STUB = <<< 'CHANGELOG'
         # Changelog
-
```
weierophinney, line 228: Do the tests that use these continue to work? I'd originally left these spaces in as they get trimmed when the nowdoc is evaluated by the engine, but I'm not sure what happens if they're omitted.

weierophinney (commented):
This is a great start, @geerteltink !

I'd also add logging in the various commands, before and/or after major activities (fetching/building changelogs, making API calls, etc.), as that will help us pinpoint where issues are occurring even if the services being called don't have logging integrated.

geerteltink:
I wasn't sure about logging in the command only or the services. I guess both is better.

stof:
When doing work through the API, it would also be great to output the response content when getting an error response from the API instead of only having the failed invariant saying "something went wrong" (or a similar generic message).

Ocramius:
That is an excellent idea: adding logging to the PSR-18 client should suffice, I suppose?

Ocramius:
Closing here: to be retried in a new PR, if needed, but the rebase work is equivalent to the dev work here :)

Issue #32, opened 2020-08-05
Document the process of creating a subkey


morozov:
> […] a **password-less** private GPG key in ASCII format, to be used for signing your releases:
>  please use a dedicated GPG subkey for this purpose.

For non-experienced in GnuPG users like myself, it will be helpful to have more detailed documentation on creating the signing subkey and exporting it to GitHub.

Ocramius:
if anyone can confirm that this is correct, please give me a thumbs up, or send a patch to include the guide in `README.md` or in a sub-dir.

Here's how I did it: 

 1. `gpg --edit-key "$YOUR_MASTER_KEY_ID"`: this opens an interactive prompt
 2. type `addkey`
 3. when asked for capabilities, select `signing` or `S`. I picked `RSA`, but that's mostly for compatibility with ancient systems
 4. type `save`

Now there is a subkey under your master key. 

## Remove password protection from the subkey

If your master key is password-protected (typical, and generally endorsed), we need to un-protect the sub-key. 

To do that, as far as I know, the only reliable way is to **delete** your master key and import only the subkey:

 1. `gpg --armor --export-secret-keys "$YOUR_MASTER_KEY_ID" > master-key.backup.asc` (hold on to this! DO NOT LOSE IT!)
 2. `gpg --export-secret-subkeys "$NEW_SUBKEY_ID!" > subkey.asc` (the exclamation mark is important, otherwise it exports everything)
 3. `gpg --delete-secret-keys "$YOUR_MASTER_KEY_ID"`
 4. `gpg --import subkey.asc`
 5. `gpg --edit-key "$NEW_SUBKEY_ID!"`: this opens an interactive prompt
 6. `passwd` (enter current password, then leave empty to remove the password requirements)
 7. `save`

## Export the secret subkey

 1. `gpg --armor --export-secret-subkeys  "$NEW_SUBKEY_ID!" > release-subkey.asc` (note: exclamation mark important)
 2. If you followed the steps above to un-protect your subkey, re-import your master key via `gpg --import master-key.backup.asc`
 3. go to `https://github.com/organizations/<your-organization>/settings/secrets` or `https://github.com/<user>/<repo>/settings/secrets` and enter the contents of `release-subkey.asc` as a new `SIGNING_SECRET_KEY` secret
 4. `gpg --armor --export "$YOUR_MASTER_KEY_ID" > public-key.asc`
[...]

PR #77, opened 2020-10-04
Update documentation around generating GPG keys to sign releases

carnage:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Adding more detail to documentation to cover areas I didn't find clear/had to lookup elsewhere while attempting to use this in a project. Adding to this as I progress through the setup process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,11 @@ to your project or organization:
 - `GIT_AUTHOR_EMAIL` - email address of the author of your releases: can be an email address of a bot account.
 - `SIGNING_SECRET_KEY` - a **password-less** private GPG key in ASCII format, to be used for signing your releases:
   please use a dedicated GPG subkey for this purpose. Unsigned releases are not supported, and won't be supported.
+  To generate a new GPG key use the following commands
+  `gpg2 --full-generate-key` Pick option 4, then type 4096 for key size, select your desired expiry. Fill out the user information and leave the password blank.
+  Once generated it will output something like `gpg: key <Key ID> marked as ultimately trusted` take a note of this Key Id to use in the next step
+  `gpg --output private.key --armor --export-secret-key <Key ID>` This will output the key to the file private.key in the correct format to paste into github. Delete the file once you are done and don't share it with anyone else
+  Optionally you can use `gpg --output public.key --armor --export <Key ID>` to export the corresponding public key. You can publish this key on your project webpage to allow users to verify your signed releases.
 - `ORGANIZATION_ADMIN_TOKEN` - if you use the file from [`examples/.github/workflows/release-on-milestone-closed.yml`](examples/.github/workflows/release-on-milestone-closed.yml),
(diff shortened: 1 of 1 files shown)


Ocramius:
Check also https://github.com/laminas/automatic-releases/issues/32 - that's the process I went through, at least :D

carnage:
I decided to go with a brand new GPG key as: 

a) My master key is stored on an encrypted flash drive, in a safe, no where near my Laptop; so effort+++
b) I perceive GPG to be a very fragile thing so any change could break my current setup
c) Was "testing" the setup so didn't feel the need to use a "real" key
d) Using a per project release key seemed like a good idea anyway.

Given your comments on #32 I'll perhaps split this into it's own document and add the details on creating a new subkey as well.

I suspect that it would be possible to make it passwordless without deleting your key; simply set the password on the master key to empty; export the sub key and restore the password to it's previous value.

Ocramius:
> a) My master key is stored on an encrypted flash drive, in a safe, no where near my Laptop; so effort+++

My master key is not even at my home, so the effort is kinda normal, FWIW.

I guess multiple GPG keys are viable in github config (https://github.com/settings/keys)

 > c) Was "testing" the setup so didn't feel the need to use a "real" key

Yeah, but docs should be about the "real" key - the actual setup workflow is indeed in #32

 > I suspect that it would be possible to make it passwordless without deleting your key; simply set the password on the master key to empty; export the sub key and restore the password to it's previous value.

Probably feasible: didn't want to change any other pre-existing subkeys on my system

carnage:
Made a few changes to this;

Pulled the GPG stuff into it's own section with sub sections for using a new key or an existing one. I used your steps from #32 as a basis for this but found a better way to remove a password from a subkey using a temp home dir so not to mess with the users current gpg setup. 

Think it's ready for merge now.

Ocramius (approved):
Excellent, thanks!

glensc:
@carnage what was your gpg version when you created these docs?

Asking as such choices are not available for me:
- https://github.com/laminas/automatic-releases/pull/102#issuecomment-766449297

Ocramius:
```
gpg (GnuPG) 2.2.4
libgcrypt 1.8.1
```

No idea if GnuPG is semver-compliant there.

glensc:
Some 3 year time delta:

Version | Date | Announce
-- | -- | --
2.2.4 | Wed Dec 20 16:36:34 CET 2017 | https://lists.gnupg.org/pipermail/gnupg-announce/2017q4/000419.html
2.2.25 | Mon Nov 23 19:00:22 CET 2020 | https://lists.gnupg.org/pipermail/gnupg-announce/2020q4/000450.html

Ocramius:
Yeah, I'm on Ubuntu 18.04 LTS, so it probably stuck to something ancient :-\

PR #97, opened 2021-01-11
Upgraded dependencies, `composer:^2` and `php:^8`, mutation tests CI improvements

Ocramius:
I was mostly annoyed by the `composer install` time in CI environments, so I decided to give this a shot and update to `php:^8` and `composer/composer:^2`.

TODO:

 * [x] add CI task that performs a `docker build`, ensuring that our build process still runs

Blockers:

 * [x] https://github.com/jwage/changelog-generator/pull/64
 * [x] https://github.com/jwage/changelog-generator/pull/65
 * [x] https://github.com/phly/keep-a-changelog/issues/96
 * [x] https://github.com/phly/keep-a-changelog/issues/97


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,21 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Simulate Docker Build"
+
+on:
+  pull_request:
+  schedule:
+    - cron: '0 2 * * *' # run daily
```
Ocramius, line 8: kinda required, since ubuntu base dependencies keep shifting.

I'd be happier if I did this properly with nixos, but I need to put in the effort to actually get there myself :-)

geerteltink:
LGTM when the dependencies are updated. This should improve the build time a lot.

@Ocramius have you thought about building the docker container itself and push that to a registry? That way the container doesn't need to be build repeatedly and would improve speed even more. I'm not sure how that would work with actions though or if we can use GitHub registry for it. Last time I checked you could not access the GitHub registry without tokens.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  docker-build:
+    name: "Simulate Docker Build"
+
+    runs-on: ${{ matrix.operating-system }}
```
geerteltink: There is no matrix configured? Why not just `"ubuntu-latest"`?

Ocramius: Good point 👍

Ocramius:
Yeah, unsure about docker builds: would be totally for it if it were globally usable 🤔

.github/workflows/docker-build.yml
Ocramius: 44344879748d773c5945a4508712c806417d0bcb

Ocramius:
Note: the added mumbojumbo around `mutation-tests.yml` is mostly to produce decent reports (and fast ones) for users that send patches to improve small sections of the code.

That ensures that we don't need to run the full mutation test suite against everything every time, if the diff is limited to some sources.

The output looks like this: 

Selection_567

Ocramius:
Just to make sure everything went smoothly, I ran this on a pre-existing tag of `ocramius/LazyMap`, and it works fine: https://github.com/Ocramius/LazyMap/runs/1739730760?check_suite_focus=true#step:4:12 (failure expected, since tag already exists)

PR #102, opened 2021-01-24
Improve documentation for gpg subkey creation

glensc:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### Description

This updates GPG section to match my system (`gpg (GnuPG) 2.2.25` from [brew]).

[brew]: https://brew.sh/

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,9 @@ First open your master key for editing (use `--list-keys` to find it):
 gpg --edit-key "<YOUR MASTER KEY ID>"
 ```
 
-Type `addkey` and select signing or s for capabilities. RSA key type is recommended for greatest compatibility.
+A signing key type RSA with 3072 bits is recommended for greatest compatibility.
+
+Type `addkey` and select a type that is for signing, you might be asked about bit size depending on your choice.
 Type `save` to persist the new subkey to your master key. Make a note of the Key ID  as you will need it in the next step.
 
 Next export the new sub key:
--- a/README.md
+++ b/README.md
@@ -46,8 +46,6 @@ First open your master key for editing (use `--list-keys` to find it):
 gpg --edit-key "<YOUR MASTER KEY ID>"
 ```
 
-A signing key type RSA with 3072 bits is recommended for greatest compatibility.
-
 Type `addkey` and select a type that is for signing, you might be asked about bit size depending on your choice.
 When deciding over key expire, avoid setting to never expire, as recommendation of key bits will change over time.
 Type `save` to persist the new subkey to your master key. Make a note of the Key ID  as you will need it in the next step.
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ in the GitHub Docs.
 
 #### Using a subkey from an existing GPG key
 
-First open your master key for editing:
+First open your master key for editing (use `--list-keys` to find it):
 
 ```bash
 gpg --edit-key "<YOUR MASTER KEY ID>"
--- a/README.md
+++ b/README.md
@@ -68,27 +68,25 @@ You can skip this if your master key is not password protected.
 To remove the password from the subkey, create an ephemeral gpg home directory:
 
 ```bash
-mkdir /tmp/gpg
+install -d -m 700 gpg-tmp
 ```
 
 Ensure that it works with gpg:
 
 ```bash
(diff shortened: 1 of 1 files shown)


glensc:
> Type `addkey` and select signing or s for capabilities. RSA key type is recommended for greatest compatibility. Type save to persist the new subkey to your master key. Make a note of the Key ID as you will need it in the next step.


I'm confused here, "s" is supposed to be key to type? i have only numeric answers:

```
gpg> addkey
Please select what kind of key you want:
   (3) DSA (sign only)
   (4) RSA (sign only)
   (5) Elgamal (encrypt only)
   (6) RSA (encrypt only)
  (14) Existing key from card
Your selection? s
Invalid selection.
```

glensc:
Next, I also get asked about bits:

```
RSA keys may be between 1024 and 4096 bits long.
What keysize do you want? (3072)
```

So what's the recommendation here for git signing in github? 

From  https://en.wikipedia.org/wiki/Key_size I read:
> Since 2015, NIST recommends a minimum of 2048-bit keys for RSA, an update to the widely-accepted recommendation of a 1024-bit minimum since at least 2002.

Ocramius:
```
Your selection? s
Invalid selection.
```

Looks like something that has changed on newer releases: certainly want a key with signing capabilities (used to be `s` for `sign`, `e` for `encrypt` and something else for certification/authentication)

 > So what's the recommendation here for git signing in github?

The default of 3072 works: bigger values lead to some overhead, but overall not noticeable. My old key is still 2048, for example.

glensc:
I've added an example commit with 2048, change to 3072?

Ocramius:
Probably better to keep it the default suggested one: that's really up to GnuPG to endorse, while we're mostly only asking for a key, rather than endorsing values to configure one (security-sensitive stuff that we shouldn't touch)

glensc:
If I attempt to remove password, I end up with a confusing error:

```
gpg> passwd
gpg: key XXXX/XXXX: error changing passphrase: No secret key

gpg> save
Key not changed so no update needed.
```

but in reality password was removed after `passwd` command, and yes, the `save` is unneeded.

README.md
```diff
@@ -40,13 +40,16 @@ in the GitHub Docs.
...
 
-First open your master key for editing:
+First open your master key for editing (use `--list-keys` to find it):
 
 ```bash
 gpg --edit-key "<YOUR MASTER KEY ID>"
 ```
 
-Type `addkey` and select signing or s for capabilities. RSA key type is recommended for greatest compatibility.
+A signing key type RSA with 3072 bits is recommended for greatest compatibility.
```
Ocramius: Discussed during other comments in the review: I'd say that the length should not be endorsed by us, since the tool already endorses one, while if we endorse it here it will become old knowledge pretty quickly.

glensc: I thought to express it's as a recommendation as is RSA recommended here. Would "at least 3072" be better?

There's (in this PR changes) also a recommendation about expiry, which addresses that the security related recommendation can become "old knowledge" pretty soon, perhaps should change the order of the text?

I think more info here is better than "go search yourself", because this project could be used by very beginner users, and it has said the gpg stuff is PITA even for novice users.

Ocramius: Would be best to not recommend a value nor provide any guidance for the key length at all: this target changes, and the first time a security researcher with a bad mood shows up we have a problem :-)

heiglandreas: For the same reasons I would not recommend any cryptosystem here. Whether people use RSA or DSA or any ElipticCurve or whatnot  is their preference and dependend on their toolset.

Ocramius: Aye: the only requirement is that the key must be usable for signing :+1:

glensc: Ok. let's remove any recommendation then, just "chose signing"

glensc:
@Ocramius I have nothing more to add here at this point.

glensc:
Another ping!

Ocramius (approved):
@glensc thanks for improving the docs here - sorry for the massive delay!

glensc:
@Ocramius sure, we have our own distraction demons. I planned to check out the "new key" section too but never got there.

PR #115, opened 2021-02-24
Reduce docker image size (and build time)

lcobucci:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This improves the build process of the docker image.

When the base images are already downloaded, it takes about 15s to build the image using BuildKit - previously it was taking 1m19s.
The image size has also been drastically reduced, it's 122Mb - previously was 390Mb.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -18,4 +18,4 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Docker Build"
-        run: "docker build ."
+        run: "DOCKER_BUILDKIT=1 docker build ."
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM composer:2 AS composer
 
-FROM ubuntu:20.04
+FROM php:8.0-alpine
 
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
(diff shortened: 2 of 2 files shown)


lcobucci:
We can compare the speed improvements on the build too: [a previous run took 1m31s](https://github.com/laminas/automatic-releases/pull/102/checks?check_run_id=1763804265) while [this one took 41s](https://github.com/laminas/automatic-releases/pull/115/checks?check_run_id=1974463684).

lcobucci:
I just noticed that we don't have the branch `1.10.x` and since this isn't a bug fix it shouldn't be merged to `1.9.x`. I'll convert it into a draft for now - to avoid accidental merges.

Dockerfile
```diff
@@ -15,33 +15,21 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
...
 # @TODO https://github.com/laminas/automatic-releases/issues/8 we skip `.git` for now, as it isn't available in the build environment
 # @TODO https://github.com/laminas/automatic-releases/issues/9 we skip `.git` for now, as it isn't available in the build environment
 #ADD .git /app/.git
-ADD bin /app/bin
-ADD src /app/src
+COPY bin /app/bin/
+COPY src /app/src/
 
-RUN composer install -a --no-dev
+RUN composer dumpautoload -a --no-dev
```
weierophinney: Why `dumpautoload` and not `install`?

lcobucci: I've tried to cover that on the [commit message](https://github.com/laminas/automatic-releases/pull/115/commits/3aba15503c21b44575cdf92f42f77ecb98ed6795). But the dependencies are already installed at this point, we just need to create the autoloader.

Ocramius (approved):
LGTM! I just need to do the branching ballet to get this released then :-)

Dockerfile
Slamdunk:
```suggestion
RUN composer dump-autoload -a --no-dev
```

See https://github.com/composer/composer/blob/e3af4ee606023079d8af222da82a5b6f36ca879f/src/Composer/Command/DumpAutoloadCommand.php#L29-L30

glensc (approved):

.github/workflows/docker-build.yml
```diff
@@ -18,4 +18,4 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Docker Build"
-        run: "docker build ."
+        run: "DOCKER_BUILDKIT=1 docker build ."
```
Ocramius, line 21: BUILDKIT FTW :muscle:

Ocramius:
Branching fixed: :ship:

lcobucci:
Thanks @Ocramius

weierophinney:
I CANNOT WAIT for my next release to see the speed improvements!

Ocramius:
@weierophinney FWIW, 43s runtime: https://github.com/laminas/automatic-releases/runs/1978710568?check_suite_focus=true

PR #116, opened 2021-02-25
Fix version comparison algorithm

### LINKED ISSUE #74: Randomly not creating next release branch
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | v1 (1.9.0?)

#### Summary

After performing the 2.9.0 release on laminas-paginator, a new release branch was not created (expected 2.10.x creation).

I'm unsure how to reproduce. What may be more useful in addressing this is to add logging wherever we can (particularly since we have a logger available now).

lcobucci:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

We were using a string-based comparison to handle this, which is rather error-prone since it considers '1.10.0' <= '1.9.0'.
This makes sure we use integers to perform the comparison in the correct hierarchy of the semver rules.

This seems to be the root cause of https://github.com/laminas/automatic-releases/issues/74.


Ocramius (approved):
LGTM :+1: 

Thanks for spotting this one, @lcobucci!

src/Git/Value/SemVerVersion.php
```diff
@@ -86,6 +86,19 @@ public function isNewMajorRelease(): bool
...
+    {
+        $comparison = $this->major <=> $other->major;
+
+        if ($comparison !== 0) {
+            return $comparison;
+        }
+
+        $comparison = $this->minor <=> $other->minor;
+
+        return $comparison !== 0 ? $comparison : $this->patch <=> $other->patch;
```
bendavies, line 102: this can all be condensed to:
```php
return [$this->major, $this->minor, $this->patch] <=> [$other->major, $other->minor, $other->patch];
```

lcobucci: @bendavies cool, didn't know. Would you like to send a patch?

bendavies: sure!

glensc: Why not use something like:
- https://github.com/composer/semver

it's been battle-tested since 2013 or so being part of the composer package.

Ocramius: Could be done, but this code was supposed to work on a very isolated scenario, so IMO, since it is tested and relatively well protected from mutants, introducing a dependency here will likely raise more issues due to the use-case scenario being widened (while it is intentionally narrow)

PR #118, opened 2021-02-25
Fix mutation test flow for branches/tags

lcobucci:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      |no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

The variable `github.base_ref` isn't set when running the flow against branches/tags, which is leading the build to fail.

This creates conditional executions to make sure the build isn't accidentally broken any more.

---

Example of successful build on a branch (after this change): https://github.com/lcobucci/automatic-releases/runs/1982270612

### DIFF
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -28,6 +28,7 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Fetch github.base_ref (for diffing)"
+        if: ${{ github.base_ref != '' }}
         run: git fetch --depth=1 origin +refs/heads/${{github.base_ref}}:refs/remotes/origin/${{github.base_ref}}
 
       - name: "Install PHP"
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
LGTM 👍 

Still note that this workflow will go away when using https://github.com/laminas/laminas-ci-matrix-action

lcobucci:
> LGTM 
> 
> Still note that this workflow will go away when using https://github.com/laminas/laminas-ci-matrix-action

@ocramius how should we distinguish between that tool and infection when building the matrix? It seems to be using the same configuration file, which is used to create the checks (as far as I understood from https://github.com/laminas/laminas-ci-matrix-container).

Ocramius:
> how should we distinguish between that tool and infection when building the matrix?

Could you rephrase this please?

lcobucci:
Sure.

We match the configuration file of the tools to know whether or not to add them to the build: https://github.com/laminas/laminas-ci-matrix-container/blob/f6941918a63718d1c39c52062c4c61d3a1afd0f4/src/create-jobs.js#L79-L175

[Infection-static-analysis-plugin](https://github.com/Roave/infection-static-analysis-plugin) uses the very same configuration file as [Infection](https://infection.github.io/guide/) but using a different entry point to run the analysis.

My question is then: how to know when to add `infection-static-analysis-plugin` or `infection` to the build matrix?

Ocramius:
Ah, you are talking about `roave/infection-static-analysis-plugin`! 

It should suffice to check if it is existing and executable in `vendor/bin/`, or part of `require-dev`, I suppose?

lcobucci:
> Ah, you are talking about `roave/infection-static-analysis-plugin`!

Yeah, that's what we use here :rofl: 

> It should suffice to check if it is existing and executable in `vendor/bin/`, or part of `require-dev`, I suppose?

I don't know... I'm not sure if the matrix calculation happens before or after the dependencies installation. We'd have to pick the approach which is less error-prone... I guess the former isn't an option since infection would also be there when `roave/infection-static-analysis-plugin` is installed, right?

Anyhow, I'll open an issue on the right repo and we can continue the discussion there :+1:

Ocramius:
> I don't know... I'm not sure if the matrix calculation happens before or after the dependencies installation. We'd have to pick the approach which is less error-prone... I guess the former isn't an option since infection would also be there when roave/infection-static-analysis-plugin is installed, right?

Agree, but it's our "best effort" for now.

 > Anyhow, I'll open an issue on the right repo and we can continue the discussion there 👍

Indeed, best way forward

PR #124, opened 2021-03-03
Modify action to build and consume container

weierophinney:
To speed things up further, we can build and release the actual container on release publication, and then have the action itself consume the image instead of the Dockerfile.

Additionally, we can force the release process to use the latest set of changes by defining a local GHA that uses the Dockerfile instead.
I've proven this works at https://github.com/weierophinney/test-local-action/, where I observed that the build happens relative to the checkout directory, and not to where the action is defined.

This will require defining two secrets in this repository:

- CONTAINER_USERNAME (username on ghcr.io)
- CONTAINER_PAT (personal authentication token with container write permissions)

Once setup, this should automatically build and push tags for the docker image, and the action will use the image to do its work.

### DIFF
--- a/.github/actions/automatic-releases/action.yml
+++ b/.github/actions/automatic-releases/action.yml
@@ -15,6 +15,6 @@ inputs:
 
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/laminas/automatic-releases:1'
+  image: '../../../Dockerfile'
   args:
     - ${{ inputs.command-name }}
--- a/.github/workflows/automatic-release.yml
+++ b/.github/workflows/automatic-release.yml
@@ -17,7 +17,7 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Release"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
         with:
           command-name: "laminas:automatic-releases:release"
         env:
@@ -27,7 +27,7 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Create Merge-Up Pull Request"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
         with:
           command-name: "laminas:automatic-releases:create-merge-up-pull-request"
         env:
@@ -37,7 +37,7 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Create and/or Switch to new Release Branch"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
         with:
           command-name: "laminas:automatic-releases:switch-default-branch-to-next-minor"
         env:
@@ -47,7 +47,7 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Bump Changelog Version On Originating Release Branch"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
         with:
           command-name: "laminas:automatic-releases:bump-changelog"
         env:
@@ -57,7 +57,7 @@ jobs:
           "GIT_AUTHOR_EMAIL": ${{ secrets.GIT_AUTHOR_EMAIL }}
 
       - name: "Create new milestones"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
         with:
           command-name: "laminas:automatic-releases:create-milestones"
         env:
--- a/.github/workflows/build-and-push-containers.yml
+++ b/.github/workflows/build-and-push-containers.yml
@@ -5,10 +5,8 @@ on:
     types: [published]
 
 jobs:
-  tags:
+  release-container:
     runs-on: ubuntu-latest
-    outputs:
-      tags: ${{ steps.tags.outputs.tags }}
     steps:
       - name: Compile tag list
         id: tags
@@ -18,12 +16,7 @@ jobs:
           MAJOR="${PREFIX}:$(echo ${TAG} | cut -d. -f1)"
           MINOR="${MAJOR}.$(echo ${TAG} | cut -d. -f2)"
           PATCH="${PREFIX}:${TAG}"
-          echo "::set-output name=tags::${MAJOR}%0A${MINOR}%0A${PATCH}"
-
-  release-container:
-    runs-on: ubuntu-latest
-    needs: [tags]
-    steps:
+          echo "::set-output name=tags::[\"${MAJOR}\",\"${MINOR}\",\"${PATCH}\"]"
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup QEMU
@@ -43,5 +36,4 @@ jobs:
           file: ./Dockerfile
           platforms: linux/amd64
           push: true
-          tags: ${{ needs.tags.outputs.tags }}
-
+          tags: ${{ join(fromJSON(steps.tags.outputs.tags), "\n") }}
--- a/.github/workflows/build-and-push-containers.yml
+++ b/.github/workflows/build-and-push-containers.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Compile tag list
         id: tags
         run: |
-          TAG=${GITHUB_REF/refs\/tags\//}
+          TAG=${GITHUB_REF#refs/tags/}
           PREFIX=ghcr.io/laminas/automatic-releases
           MAJOR="${PREFIX}:$(echo ${TAG} | cut -d. -f1)"
           MINOR="${MAJOR}.$(echo ${TAG} | cut -d. -f2)"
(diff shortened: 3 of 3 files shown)


weierophinney:
I've added the secrets at this point. After first build, we'll need to make the container public, and then associate it with this repository; I'll go ahead and do that when released.

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - name: Compile tag list
+        id: tags
+        run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
+          PREFIX=ghcr.io/laminas/automatic-releases
+          MAJOR="${PREFIX}:$(echo ${TAG} | cut -d. -f1)"
+          MINOR="${MAJOR}.$(echo ${TAG} | cut -d. -f2)"
+          PATCH="${PREFIX}:${TAG}"
+          echo "::set-output name=tags::${MAJOR}%0A${MINOR}%0A${PATCH}"
```
weierophinney: GHA replaces `%0A` with `\n`, making these line-delimited tags. This creates a list that looks like:

```text
ghcr.io/laminas/automatic-releases:1
ghcr.io/laminas/automatic-releases:1.11
ghcr.io/laminas/automatic-releases:1.11.0
```

which is then passed as the variable "tags" to the next step.

.github/actions/automatic-releases/action.yml
```diff
@@ -0,0 +1,20 @@
...
+       * `laminas:automatic-releases:release`
+       * `laminas:automatic-releases:create-merge-up-pull-request`
+       * `laminas:automatic-releases:switch-default-branch-to-next-minor`
+    required: true
+
+runs:
+  using: 'docker'
+  image: '../../../Dockerfile'
+  args:
+    - ${{ inputs.command-name }}
```
weierophinney, line 20: What I discovered is that while the image is in a directory relative to this one, the actual build happens relative to `$GITHUB_WORKSPACE`, so everything works correctly. By doing this, we get the benefits of whatever changes we've made for this release... when creating the release.

action.yml
```diff
@@ -15,6 +15,6 @@ inputs:
 
 runs:
   using: 'docker'
-  image: 'Dockerfile'
+  image: 'docker://ghcr.io/laminas/automatic-releases:1'
```
weierophinney, line 18: This is the magic: the action now uses an image on ghcr.io, which means only a pull operation is required, and no build operations!

.github/workflows/automatic-release.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Release"
-        uses: "./"
+        uses: "./.github/actions/automatic-releases/"
```
weierophinney, line 20: And this part ensures we use that local action, which builds from the Dockerfile directly instead of using the image.

.github/workflows/build-and-push-containers.yml
glensc: This is probably not the best idea to rely on undocumented behavior. Seems like an implementation detail that may behave differently in the future.

However, they have documented to pass as json, if you need something like this:

```yaml
name: build
on: push
jobs:
  job1:
    runs-on: ubuntu-latest
    outputs:
      matrix: ${{ steps.set-matrix.outputs.matrix }}
    steps:
    - id: set-matrix
      run: echo "::set-output name=matrix::{\"include\":[{\"project\":\"foo\",\"config\":\"Debug\"},{\"project\":\"bar\",\"config\":\"Release\"}]}"
  job2:
    needs: job1
    runs-on: ubuntu-latest
    strategy:
      matrix: ${{fromJSON(needs.job1.outputs.matrix)}}
    steps:
    - run: build
```

- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-returning-a-json-object

weierophinney: @glensc This is actually the approach recommended by the GHA team, and I discovered it here: https://github.community/t/set-output-truncates-multiline-strings/16852/3 (they are indicating that they transform ANSI escape sequences emitted in outputs).

A later note in that thread indicates that it's undocumented, but that they will be updating the docs to reflect this.

weierophinney: I've done some thinking on this, and decided to go with the JSON approach; I'm now doing:

```yaml
with:
  tags: ${{ join(fromJSON(steps.tags.outputs.tags), "\n") }}
```

(as they "tags" value is expected to be multiline).

This should address any concerns about ANSI escape sequences, and ensure we are forwards-compatible with any changes GHA makes.

weierophinney: Update: using "," as the separator, as CSV is supported for that value, and the `"\n"` value was causing issues with the GHA expressions.

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release-container:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Compile tag list
```
bendavies, line 11: you could use https://github.com/crazy-max/ghaction-docker-meta here

weierophinney: True - but this is fast and doesn't require any additional overhead as it's using tooling already in the environment.

Thanks for the link, though - I'd not seen that one!

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: tags
+        run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
+          PREFIX=ghcr.io/laminas/automatic-releases
+          MAJOR="${PREFIX}:$(echo ${TAG} | cut -d. -f1)"
+          MINOR="${MAJOR}.$(echo ${TAG} | cut -d. -f2)"
+          PATCH="${PREFIX}:${TAG}"
+          echo "::set-output name=tags::[\"${MAJOR}\",\"${MINOR}\",\"${PATCH}\"]"
+      - name: Checkout
+        uses: actions/checkout@v2
```
Ocramius, line 22: Not super-important, but some spacing (blank lines around blocks) would go a long way with improving readability

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.CONTAINER_USERNAME }}
```
Ocramius: These secrets should be prefixed, to make it clear they are local to our build

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,39 @@
...
+          password: ${{ secrets.CONTAINER_PAT }}
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64
+          push: true
+          tags: ${{ join(fromJSON(steps.tags.outputs.tags), ",") }}
```
Ocramius, line 44: So we push each image as 3 tags? Ok for me, hard to read/follow if not versed in github actions (requires comments around it)

Ocramius (approved):
Besides minor improvement suggestions, LGTM :+1:

.github/workflows/build-and-push-containers.yml
weierophinney: Yes - actual release version, minor release version, major release version.

Rationale is that if we change something in a new minor release that causes issues for somebody, they can pin to the previous minor. Alternately, they can pin to a specific "known good" release. (There are existing actions for creating tags that do the same; I chose to do it using standard *nix tooling, as it introduces zero additional overhead.)

.github/workflows/build-and-push-containers.yml
weierophinney: Done!

.github/workflows/build-and-push-containers.yml
```diff
@@ -0,0 +1,39 @@
...
+    types: [published]
+
+jobs:
+  release-container:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Compile tag list
+        id: tags
+        run: |
+          TAG=${GITHUB_REF/refs\/tags\//}
```
glensc:
```suggestion
          TAG=${GITHUB_REF#refs/tags/}
```

I'm sure I've already suggested this elsewhere.

.github/workflows/build-and-push-containers.yml
weierophinney: Done; used `AUTOMATIC_RELEASES_` as the prefix.

.github/workflows/build-and-push-containers.yml
weierophinney: Done.

PR #129, opened 2021-04-11
Feature: Twitter release announcement

### LINKED ISSUE #24: Automate release announcement on twitter
### Feature Request

I find myself always having to announce a release on twitter after closing a milestone.

While that's nice (added visibility), it is also something that could maybe be automated, although unsure how to formulate a meaningful tweet from it.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

ghostwriter:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| RFC           | yes

### Description

This PR attempts to resolve #24

- Add `abraham/twitteroauth`

- Add support for Twitter EnvironmentVariables

- Update example workflows

Tweet will be extracted from the milestone, if code block matches:
> \``` tweet
my announcement tweet goes here.
\```

If one is not found, it will create a default tweet.
```
Released: {repository} {version} https://github.com/{repository}/releases/tag/{version}
```


src/Twitter/Value/Tweet.php
```diff
@@ -0,0 +1,50 @@
...
+        $version    = $event->version()->fullReleaseName();
+        /** @psalm-var non-empty-string $content*/
+        $content = str_replace([
+            '{repository}',
+            '{version}',
+            '{link}',
+        ], [
+            $repository,
+            $version,
+            'https://github.com/laminas/laminas-navigation/releases/tag/' . $version,
```
froschdesign: The URL looks wrong here.

Ocramius:
I jumped in waaaay too late on this: I think this should perhaps be an `on: tag` feature, rather than a manually triggered command that operates on a milestone.

EDIT: sorry, tabbed to `close and comment` instead of `comment` -.-

src/Environment/EnvironmentVariables.php
```diff
@@ -88,6 +108,10 @@ public static function fromEnvironment(ImportGpgKeyFromString $importKey): self
             self::getenv('GITHUB_EVENT_PATH'),
             self::getenv('GITHUB_WORKSPACE'),
             self::getenvWithFallback('LOG_LEVEL', 'INFO'),
+            self::getenv('TWITTER_CONSUMER_API_KEY'),
```
Ocramius: This makes configuring these API keys kinda required, which is a problem.

Would be best to have them completely separately configured, perhaps?

Also, are 4 secrets really needed? Sounds complex :thinking:

src/Twitter/CreateTweetThroughApiCall.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function __invoke(MilestoneClosedEvent $event): UriInterface
+    {
+        $tweet = Tweet::fromMilestoneClosedEvent($event);
+
+        $response = $this->statusesUpdate($tweet);
+
+        $responseObject = $response->toValue();
+        Assert::isInstanceOf($responseObject, stdClass::class);
```
Ocramius, line 34: Urgh, is our API client this bad? :S

src/Twitter/Value/Tweet.php
```diff
@@ -0,0 +1,46 @@
...
+
+use Laminas\AutomaticReleases\Github\Event\MilestoneClosedEvent;
+
+use function str_replace;
+
+/** @psalm-immutable */
+final class Tweet
+{
+    /** @psalm-var non-empty-string */
+    private const TEMPLATE = 'Released: {repository} {version} https://github.com/{repository}/releases/tag/{version}';
```
Ocramius, line 15: I wonder if there could be a short description, extracted from the milestone, to tweet some text with it?

src/Twitter/CreateTweetThroughApiCall.php
weierophinney: We're doing the same as most API clients and just running json_decode() on the response payload, because it keeps us future-proof. 😄

src/Environment/EnvironmentVariables.php
weierophinney: Twitter oauth1 crap.

src/Twitter/CreateTweetThroughApiCall.php
Ocramius: Did you just tell me "yes, our API client is that bad!"?

glensc: At least it's not terrible, so it's good that it's bad, so at least it's not terrible :D

weierophinney: Let's just say it's of roughly equal quality to every other API client out there. :stuck_out_tongue:

weierophinney:
> I jumped in waaaay too late on this: I think this should perhaps be an `on: tag` feature, rather than a manually triggered command that operates on a milestone.

I'm actually thinking this too, especially since there are already actions out there that can do this.

As an example, the [Send Tweet Action](https://github.com/marketplace/actions/send-tweet-action) could be configured into a workflow:

```yaml
name: Send a Tweet
on:
  release:
    types: [published]
jobs:
  tweet:
    runs-on: ubuntu-latest
    steps:
      - id: notes
        run: |
          MESSAGE="Just released: ${GITHUB_REPOSITORY} ${GITHUB_REF}: https://github.com/${GITHUB_REPOSITORY}/releases/tag/${GITHUB_REF}"
          echo "::set-output name=status::${MESSAGE}"

      - uses: ethomson/send-tweet-action@v1
        with:
          status: ${{ steps.notes.outputs.status }}
          consumer-key: ${{ secrets.TWITTER_CONSUMER_API_KEY }}
          consumer-secret: ${{ secrets.TWITTER_CONSUMER_API_SECRET }}
          access-token: ${{ secrets.TWITTER_ACCESS_TOKEN }}
          access-token-secret: ${{ secrets.TWITTER_ACCESS_TOKEN_SECRET }}
```

This can then be dropped into a separate workflow file. This approach makes it completely opt-in, which I like, and allows customizing the status message however you want.

PR #138, opened 2021-05-21
replace `thecodingmachine/safe`, `webmozart/assert`, and `symfony/process` by `azjezz/psl`

azjezz:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Similar to https://github.com/Roave/BackwardCompatibilityCheck/pull/306, this PR replaces `thecodingmachine/safe`, `webmozart/assert`, and `symfony/process` by `azjezz/psl`.

This reduces the amount of dependencies, and relies on a better type assertion component provided by PSL.

Psl also ships with functions that replace PHP builtin to provide a better error handling, and consistency.

`symfony/process` is not made absolute by `Psl\Shell\execute`, as `execute` function is created as a replacement for `exec` builtin function, however, it provides enough functionality ( working directory, argument escaping, environment variables .. etc ) to replace the usage of `symfony/process` within this library ( if more features are needed later on, it makes sense to bring `symfony/process` back ).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,12 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
 
 WORKDIR /app
 
-RUN apk add --no-cache git gnupg libzip \
+RUN apk add --no-cache git gnupg libzip icu-dev \
     && apk add --no-cache --virtual .build-deps libzip-dev \
     && docker-php-ext-install zip \
+    && docker-php-ext-install bcmath \
+    && docker-php-ext-configure intl \
+    && docker-php-ext-install intl \
     && apk del .build-deps
 
 COPY composer.* /app/
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -29,5 +29,6 @@
     </issueHandlers>
     <plugins>
         <pluginClass class="Psalm\PhpUnitPlugin\Plugin"/>
+        <pluginClass class="Psl\Psalm\Plugin"/>
     </plugins>
 </psalm>
--- a/src/Application/Command/CreateMergeUpPullRequest.php
+++ b/src/Application/Command/CreateMergeUpPullRequest.php
@@ -13,13 +13,13 @@ use Laminas\AutomaticReleases\Github\Api\GraphQL\Query\GetGithubMilestone;
 use Laminas\AutomaticReleases\Github\Api\V3\CreatePullRequest;
 use Laminas\AutomaticReleases\Github\CreateReleaseText;
 use Laminas\AutomaticReleases\Github\Event\Factory\LoadCurrentGithubEvent;
+use Psl;
+use Psl\Filesystem;
+use Psl\SecureRandom;
+use Psl\Str;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Webmozart\Assert\Assert;
-
-use function sprintf;
-use function uniqid;
 
 final class CreateMergeUpPullRequest extends Command
 {
@@ -59,7 +59,7 @@ final class CreateMergeUpPullRequest extends Command
         $event          = $this->loadGithubEvent->__invoke();
         $repositoryPath = $this->variables->githubWorkspacePath();
 
-        Assert::directory($repositoryPath . '/.git');
+        Psl\invariant(Filesystem\is_directory($repositoryPath . '/.git'), 'Workspace is not a GIT repository.');
 
         $this->fetch->__invoke(
             $event->repository()
@@ -74,7 +74,7 @@ final class CreateMergeUpPullRequest extends Command
         $mergeUpTarget  = $mergeCandidates->branchToMergeUp($releaseVersion);
 
         if ($mergeUpTarget === null) {
-            $output->writeln(sprintf(
+            $output->writeln(Str\format(
                 'No merge-up candidate for release %s - skipping pull request creation',
                 $releaseVersion->fullReleaseName()
             ));
@@ -82,16 +82,14 @@ final class CreateMergeUpPullRequest extends Command
             return 0;
         }
 
-        Assert::notNull(
-            $releaseBranch,
-            sprintf('No valid release branch found for version %s', $releaseVersion->fullReleaseName())
-        );
+        Psl\invariant($releaseBranch !== null, Str\format('No valid release branch found for version %s', $releaseVersion->fullReleaseName()));
 
         $mergeUpBranch = BranchName::fromName(
             $releaseBranch->name()
             . '-merge-up-into-'
             . $mergeUpTarget->name()
-            . uniqid('_', true) // This is to ensure that a new merge-up pull request is created even if one already exists
+            . '_'
+            . SecureRandom\string(8) // This is to ensure that a new merge-up pull request is created even if one already exists
         );
 
         $releaseNotes = $this->createReleaseText->__invoke(
--- a/src/Application/Command/CreateMilestones.php
+++ b/src/Application/Command/CreateMilestones.php
@@ -41,14 +41,12 @@ final class CreateMilestones extends Command
         return 0;
     }
 
-    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): bool
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): void
     {
         try {
             ($this->createMilestone)($repositoryName, $version);
-        } catch (CreateMilestoneFailed $e) {
-            return false;
+        } catch (CreateMilestoneFailed) {
+            return;
         }
-
-        return true;
     }
 }
--- a/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
+++ b/src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
@@ -11,10 +11,9 @@ use Laminas\AutomaticReleases\Git\Value\BranchName;
 use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
 use Phly\KeepAChangelog\Bump\ChangelogBump;
+use Psl\Str;
+use Psl\Type;
 use Psr\Log\LoggerInterface;
-use Webmozart\Assert\Assert;
-
-use function sprintf;
 
 class BumpAndCommitChangelogVersionViaKeepAChangelog implements BumpAndCommitChangelogVersion
 {
@@ -62,16 +61,15 @@ class BumpAndCommitChangelogVersionViaKeepAChangelog implements BumpAndCommitCha
 
         ($this->checkoutBranch)($repositoryDirectory, $sourceBranch);
 
-        $changelogFile = sprintf('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
+        $changelogFile = Str\format('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
         $versionString = $version->fullReleaseName();
         $bumper        = new ChangelogBump($changelogFile);
-        $newVersion    = $bumper->$bumpType($versionString);
+        $newVersion    = Type\non_empty_string()->assert($bumper->$bumpType($versionString));
 
-        Assert::stringNotEmpty($newVersion);
         $bumper->updateChangelog($newVersion);
 
-        $message = sprintf(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion);
-        Assert::notEmpty($message);
+        $message = Type\non_empty_string()
+            ->assert(Str\format(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion));
 
         ($this->commitFile)(
             $repositoryDirectory,
@@ -83,7 +81,7 @@ class BumpAndCommitChangelogVersionViaKeepAChangelog implements BumpAndCommitCha
 
         ($this->push)($repositoryDirectory, $sourceBranch->name());
 
-        $this->logger->info(sprintf(
+        $this->logger->info(Str\format(
             'BumpAndCommitChangelogVersion: bumped %s to version %s in branch %s',
             self::CHANGELOG_FILE,
             $newVersion,
--- a/src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
+++ b/src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
@@ -10,10 +10,9 @@ use Laminas\AutomaticReleases\Git\Push;
 use Laminas\AutomaticReleases\Git\Value\BranchName;
 use Laminas\AutomaticReleases\Git\Value\SemVerVersion;
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
+use Psl\Str;
+use Psl\Type;
 use Psr\Log\LoggerInterface;
-use Webmozart\Assert\Assert;
-
-use function sprintf;
 
 final class CommitReleaseChangelogViaKeepAChangelog implements CommitReleaseChangelog
 {
@@ -71,13 +70,13 @@ final class CommitReleaseChangelogViaKeepAChangelog implements CommitReleaseChan
 
         ($this->checkoutBranch)($repositoryDirectory, $sourceBranch);
 
-        $changelogFile = sprintf('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
-        Assert::stringNotEmpty($changelogFile);
+        $changelogFile = Type\non_empty_string()
+            ->assert(Str\format('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE));
 
         $releaseNotes::writeChangelogFile($changelogFile, $releaseNotes);
 
-        $message = sprintf(self::COMMIT_TEMPLATE, $version->fullReleaseName(), self::CHANGELOG_FILE);
-        Assert::notEmpty($message);
+        $message = Type\non_empty_string()
+            ->assert(Str\format(self::COMMIT_TEMPLATE, $version->fullReleaseName(), self::CHANGELOG_FILE));
 
         ($this->commitFile)(
             $repositoryDirectory,
--- a/src/Environment/EnvironmentVariables.php
+++ b/src/Environment/EnvironmentVariables.php
@@ -6,14 +6,33 @@ namespace Laminas\AutomaticReleases\Environment;
 
 use Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromString;
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
-use Webmozart\Assert\Assert;
-
-use function getenv;
-use function sprintf;
+use Psl;
+use Psl\Env;
+use Psl\Iter;
+use Psl\Str;
 
 /** @psalm-immutable */
 class EnvironmentVariables implements Variables
 {
+    private const LOG_LEVELS = [
+        '100',
+        '200',
+        '250',
+        '300',
+        '400',
+        '500',
+        '550',
+        '600',
+        'DEBUG',
+        'INFO',
+        'NOTICE',
+        'WARNING',
+        'ERROR',
+        'CRITICAL',
+        'ALERT',
+        'EMERGENCY',
+    ];
+
     /** @psalm-var non-empty-string */
     private string $githubToken;
     private SecretKeyId $signingSecretKey;
@@ -52,29 +71,13 @@ class EnvironmentVariables implements Variables
         $this->githubEventPath  = $githubEventPath;
         $this->workspacePath    = $workspacePath;
 
-        Assert::inArray(
-            $logLevel,
-            [
-                '100',
-                '200',
-                '250',
-                '300',
-                '400',
-                '500',
-                '550',
-                '600',
-                'DEBUG',
-                'INFO',
-                'NOTICE',
-                'WARNING',
-                'ERROR',
-                'CRITICAL',
-                'ALERT',
-                'EMERGENCY',
-            ],
+        /** @psalm-suppress ImpureFunctionCall the {@see \Psl\Iter\contains()} API is conditionally pure */
+        Psl\invariant(
+            Iter\contains(self::LOG_LEVELS, $logLevel),
             'LOG_LEVEL env MUST be a valid monolog/monolog log level constant name or value;'
             . ' see https://github.com/Seldaek/monolog/blob/master/doc/01-usage.md#log-levels'
         );
+
         $this->logLevel = $logLevel;
     }
 
@@ -97,9 +100,9 @@ class EnvironmentVariables implements Variables
      */
     private static function getenv(string $key): string
     {
-        $value = getenv($key);
+        $value = Env\get_var($key);
 
-        Assert::stringNotEmpty($value, sprintf('Could not find a value for environment variable "%s"', $key));
+        Psl\invariant($value !== null && $value !== '', Str\format('Could not find a value for environment variable "%s"', $key));
 
         return $value;
     }
@@ -110,9 +113,9 @@ class EnvironmentVariables implements Variables
      */
     private static function getenvWithFallback(string $key, string $default): string
     {
-        $value = getenv($key);
+        $value = Env\get_var($key);
 
-        return $value === false || $value === '' ? $default : $value;
+        return $value === null || $value === '' ? $default : $value;
     }
 
     public function githubToken(): string
--- a/src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
+++ b/src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace Laminas\AutomaticReleases\Git;
 
 use Laminas\AutomaticReleases\Environment\EnvironmentVariables;
+use Psl\Shell;
 use Psr\Http\Message\UriInterface;
-use Symfony\Component\Process\Process;
 
 final class FetchAndSetCurrentUserByReplacingCurrentOriginRemote implements Fetch
 {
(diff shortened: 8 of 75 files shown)


src/Application/Command/CreateMilestones.php
```diff
@@ -41,14 +41,12 @@ public function execute(InputInterface $input, OutputInterface $output): int
         return 0;
     }
 
-    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): bool
+    private function createMilestoneIfNotExists(RepositoryName $repositoryName, SemVerVersion $version): void
```
azjezz, line 44: i believe `bool` was used to avoid the `UnusedMethodCall` psalm issue, but i think it's better to suppress it, can be reverted.

froschdesign:
@azjezz 
> This reduces the amount of dependencies…

The number of dependent libraries is reduced, but the dependency to one library is increased. Because function calls that were previously based on classic PHP are replaced.
Therefore, I think that this argument does not work here.

(_My statement refers only to this one argument and does not make any technical assessment!_)

azjezz:
> The number of dependent libraries is reduced, but the dependency to one library is increased

Yes, but that's -3 vs +1. so we are still down 2.

> Because function calls that were previously based on classic PHP are replaced.

see https://github.com/Roave/BackwardCompatibilityCheck/pull/306#discussion_r634992602 and https://github.com/Roave/BackwardCompatibilityCheck/pull/306#discussion_r634975147 on why builtin functions were replaced.

froschdesign:
@azjezz 
> …on why builtin functions were replaced.

Again: I have not made any technical assessment of the individual functions, I have only said that by replacing classical functions, the dependence to one library is increased. That is not a criticism or evaluation on `Str\format` or `Str\trim_right` itself!
And for that reason, the argument doesn't work for me – and I only mean this one argument!

psalm.xml.dist
```diff
@@ -26,8 +26,12 @@
                 <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::with"/>
             </errorLevel>
         </InternalMethod>
+
+        <UnusedMethodCall errorLevel="suppress" />
```
Ocramius: We should really avoid suppressing these overall

psalm.xml.dist
```diff
@@ -26,8 +26,12 @@
...
+        <UnusedVariable errorLevel="suppress" />
```
Ocramius: We should really avoid suppressing these overall

src/Application/Command/CreateMergeUpPullRequest.php
```diff
@@ -74,24 +74,22 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
-            $releaseBranch,
-            sprintf('No valid release branch found for version %s', $releaseVersion->fullReleaseName())
-        );
+        Psl\invariant($releaseBranch !== null, Str\format('No valid release branch found for version %s', $releaseVersion->fullReleaseName()));
 
         $mergeUpBranch = BranchName::fromName(
             $releaseBranch->name()
             . '-merge-up-into-'
             . $mergeUpTarget->name()
-            . uniqid('_', true) // This is to ensure that a new merge-up pull request is created even if one already exists
```
Ocramius, line 94: The comment should be preserved

src/Application/Command/CreateMilestones.php
Ocramius: Let's suppress it locally, rather than suppressing it globally in `psalm.xml`

Dockerfile
```diff
@@ -15,9 +15,12 @@ LABEL "maintainer"="https://github.com/laminas/technical-steering-committee/"
 
 WORKDIR /app
 
-RUN apk add --no-cache git gnupg libzip \
+RUN apk add --no-cache git gnupg libzip icu-dev \
     && apk add --no-cache --virtual .build-deps libzip-dev \
     && docker-php-ext-install zip \
+    && docker-php-ext-install bcmath \
+    && docker-php-ext-configure intl \
```
Ocramius, line 22: Are Intl/ICU avoidable? None of the contents of this library really care about i18n

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
```diff
@@ -62,16 +61,15 @@ public function __invoke(
 
         ($this->checkoutBranch)($repositoryDirectory, $sourceBranch);
 
-        $changelogFile = sprintf('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
+        $changelogFile = Str\format('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
         $versionString = $version->fullReleaseName();
         $bumper        = new ChangelogBump($changelogFile);
-        $newVersion    = $bumper->$bumpType($versionString);
+        $newVersion    = Type\non_empty_string()->assert($bumper->$bumpType($versionString));
```
Ocramius, line 67: IMO coerce? Or does `assert()` return anything?

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
```diff
@@ -62,16 +61,15 @@ public function __invoke(
...
 
-        Assert::stringNotEmpty($newVersion);
         $bumper->updateChangelog($newVersion);
 
-        $message = sprintf(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion);
-        Assert::notEmpty($message);
+        $message = Type\non_empty_string()
+            ->assert(Str\format(self::COMMIT_TEMPLATE, $newVersion, self::CHANGELOG_FILE, $newVersion));
```
Ocramius, line 72: `coerce()`?

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
```diff
@@ -71,13 +70,13 @@ public function __invoke(
 
         ($this->checkoutBranch)($repositoryDirectory, $sourceBranch);
 
-        $changelogFile = sprintf('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE);
-        Assert::stringNotEmpty($changelogFile);
+        $changelogFile = Type\non_empty_string()
+            ->assert(Str\format('%s/%s', $repositoryDirectory, self::CHANGELOG_FILE));
```
Ocramius, line 74: `coerce`?

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
```diff
@@ -71,13 +70,13 @@ public function __invoke(
...
 
         $releaseNotes::writeChangelogFile($changelogFile, $releaseNotes);
 
-        $message = sprintf(self::COMMIT_TEMPLATE, $version->fullReleaseName(), self::CHANGELOG_FILE);
-        Assert::notEmpty($message);
+        $message = Type\non_empty_string()
+            ->assert(Str\format(self::COMMIT_TEMPLATE, $version->fullReleaseName(), self::CHANGELOG_FILE));
```
Ocramius, line 79: `coerce`?

src/Environment/EnvironmentVariables.php
```diff
@@ -110,9 +113,9 @@ private static function getenv(string $key): string
      */
     private static function getenvWithFallback(string $key, string $default): string
     {
-        $value = getenv($key);
+        $value = Env\get_var($key);
 
-        return $value === false || $value === '' ? $default : $value;
+        return empty($value) ? $default : $value;
```
Ocramius: `empty()` should really never be used, since `'0'` also matches

src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
```diff
@@ -21,19 +21,14 @@ public function __invoke(
...
-            ->mustRun();
-
-        (new Process(['git', 'config', 'user.email', $this->variables->gitAuthorEmail()], $repositoryRootDirectory))
-            ->mustRun();
-
-        (new Process(['git', 'config', 'user.name', $this->variables->gitAuthorName()], $repositoryRootDirectory))
-            ->mustRun();
+        try {
+            Shell\execute('git', ['remote', 'rm', 'origin'], $repositoryRootDirectory);
+        } catch (Shell\Exception\FailedExecutionException) {
```
Ocramius, line 26: Possibly to be tested (mutant should occur here)

src/Git/Value/BranchName.php
```diff
@@ -23,12 +20,14 @@ private function __construct(string $name)
         $this->name = $name;
     }
 
-    /** @psalm-pure */
+    /**
+     * @pure
+     * @psalm-suppress ImpureFunctionCall the {@see \Psl\Type\non_empty_string()} API is pure by design
```
Ocramius, line 25: Is this function not declared pure? It should be :thinking:

src/Git/Value/BranchName.php
```diff
@@ -37,32 +36,28 @@ public function name(): string
...
+        return Regex\matches($this->name, '/^(v)?\d+\\.\d+(\\.x)?$/');
     }
 
     /**
      * @return array<int, int>
      *
      * @psalm-return array{0: int, 1: int}
      *
-     * @psalm-suppress ImpureFunctionCall the {@see \Safe\preg_match()} API is pure by design
+     * @psalm-suppress ImpureFunctionCall the {@see \Psl\Type\int()} and {@see \Psl\Type\shape()} APIs are pure by design
```
Ocramius, line 49: Is this function not declared pure? It should be :thinking:

src/Git/Value/SemVerVersion.php
```diff
@@ -24,16 +23,14 @@ private function __construct(int $major, int $minor, int $patch)
...
     public static function fromMilestoneName(string $name): self
     {
-        Assert::notEmpty($name);
-        Assert::regex($name, '/^(v)?\\d+\\.\\d+\\.\\d+$/');
-
-        preg_match('/(\\d+)\\.(\\d+)\\.(\\d+)/', $name, $matches);
+        Psl\invariant(Regex\matches($name, '/^(v)?\\d+\\.\\d+\\.\\d+$/'), 'Milestone name is malformed.');
 
-        Assert::isList($matches);
+        /** @var array{1: string, 2: string, 3: string} $matches */
```
Ocramius: Can coercion improve here perhaps? :thinking:

.github/workflows/mutation-tests.yml
```diff
@@ -65,7 +65,9 @@ jobs:
           vendor/bin/roave-infection-static-analysis-plugin \
             --ignore-msi-with-no-mutations \
             --git-diff-filter=AM \
-            --git-diff-base=origin/${{ github.base_ref }}
+            --git-diff-base=origin/${{ github.base_ref }} \
+            --min-covered-msi=100 \
```
Ocramius: These can/should be set in `infection.json.dist` (that way, you get same results also locally)

src/Github/Api/V3/CreatePullRequestThroughApiCall.php
```diff
@@ -63,16 +62,14 @@ public function __invoke(
...
-        Assert::greaterThanEq($response->getStatusCode(), 200);
-        Assert::lessThanEq($response->getStatusCode(), 299);
-
-        $responseData = json_decode($responseBody, true);
-
-        Assert::isMap($responseData);
-        Assert::keyExists($responseData, 'url');
+        Json\typed($responseBody, Type\shape([
+            'url' => Type\mixed(),
+        ], allow_unknown_fields: true));
```
Ocramius: please avoid this - only confusing, would prefer just the boolean

src/Github/CreateReleaseTextThroughChangelog.php
```diff
@@ -109,14 +97,13 @@ private function normalizeChangelog(string $changelog, string $version): string
...
-        $lines         = explode("\n", trim($changelog));
+        $lines         = Str\split(Str\trim($changelog), "\n");
         $linesToRemove = [];
         foreach ($lines as $i => $line) {
-            $matches = [];
-
             // Does the line match one of the delimiter line types? If so,
             // capture that in $matches.
-            if (! preg_match('/^(?P<delim>-{3,}|={3,})$/', $line, $matches)) {
+            $matches = Regex\first_match($line, '/^(?P<delimiter>-{3,}|={3,})$/', Regex\capture_groups(['delimiter']));
```
Ocramius, line 105: Confused by this: `$matches` is a `first_match()` here, but below we do `$matches['delimiter'][0];`?

src/Github/CreateReleaseTextViaKeepAChangelog.php
```diff
@@ -89,10 +87,8 @@ public function canCreateReleaseText(
                     $semVerVersion->fullReleaseName()
                 );
 
-            Assert::notEmpty($changelog);
-
-            return true;
-        } catch (ExceptionInterface | InvalidArgumentException $e) {
+            return ! Str\is_empty($changelog);
+        } catch (ExceptionInterface | InvalidArgumentException) {
```
Ocramius: I think `InvalidArgumentException` no longer occurs here

.github/workflows/mutation-tests.yml
```diff
@@ -65,14 +65,16 @@ jobs:
...
+            --min-msi=97
         env:
           INFECTION_BADGE_API_KEY: ${{ secrets.INFECTION_BADGE_API_KEY }}
           STRYKER_DASHBOARD_API_KEY: ${{ secrets.STRYKER_DASHBOARD_API_KEY }}
 
       - name: "Infection"
         if: ${{ github.base_ref == '' }}
-        run: vendor/bin/roave-infection-static-analysis-plugin --ignore-msi-with-no-mutations
+        run: vendor/bin/roave-infection-static-analysis-plugin --ignore-msi-with-no-mutations --min-covered-msi=100 --min-msi=97
```
Ocramius: These can/should be set in infection.json.dist (that way, you get same results also locally)

src/Gpg/ImportGpgKeyFromStringViaTemporaryFile.php
```diff
@@ -4,35 +4,27 @@
...
-        file_put_contents($keyFileName, $keyContents);
-
-        $output = (new Process(['gpg', '--import', $keyFileName]))
-            ->mustRun()
-            ->getErrorOutput();
+        $keyFileName = Filesystem\create_temporary_file(Env\temp_dir(), 'imported-key');
+        Filesystem\write_file($keyFileName, $keyContents);
 
-        Assert::regex($output, '/key\\s+([A-F0-9]+):\\s+secret\\s+key\\s+imported/im');
+        // redirect output from STDERR to STDOUT since Shell\execute only returns STDOUT content.
```
Ocramius, line 20: Is there a test for this redirect? Escaping `2>&1` is kinda fiddly

test/unit/Application/BumpChangelogForReleaseBranchTest.php
```diff
@@ -17,27 +17,23 @@
...
-use function mkdir;
-use function sys_get_temp_dir;
-use function tempnam;
-use function unlink;
-
 class BumpChangelogForReleaseBranchTest extends TestCase
 {
     /** @var Variables&MockObject */
-    private $environment;
+    private Variables | MockObject $environment;
```
Ocramius: These should not contain `|`, as that is too wide.

Ocramius: This applies to all tests involved: a `s/| MockObject/` should be performed

Ocramius (changes requested):
Stellar improvement! Very happy with the changes so far, so the comments are mostly about small details, and some type-related issues in the test suite.

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
azjezz: assert returns as well.

difference:

coerce: given `$value` of `mixed` it will try to convert `$value` to `T` and return `T`, otherwise throws.

assert: given `$value` of `mixed` it will assert that `$value` is already `T` and return it, otherwise throws.

e.g: `bool()->coerce(1); // true`, `bool()->assert(1); // throws`.

azjezz: in this specific scenario, `coerce(45)` will return`"45"` ( since it can be converted to a non-empty-string ), `assert(45);` will throw  .

so you should use assert whenever possible.

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
azjezz: same as above.

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
azjezz: same as above.

src/Changelog/CommitReleaseChangelogViaKeepAChangelog.php
azjezz: same as above.

src/Environment/EnvironmentVariables.php
azjezz: oh, my bad :)

src/Git/Value/BranchName.php
azjezz: yea, i figured i will do it in 1.8, but assert() and coerce will always be non-pure ( conditional purity )

src/Github/CreateReleaseTextThroughChangelog.php
azjezz: we are accessing an offset of a string, not array 🤷‍♂️ can use slice there if it makes things more clear, but it was always that way.

Dockerfile
azjezz: psl requires intl and bcmath, intl ext requires icu.

psl doesn't do checks like `if(extension_loaded('intl'))`, but requires all extensions in `composer.json`.

azjezz: the fact that you didn't notice it before proves to me that it's the right things to do: people most likely have them locally.

src/Git/Value/SemVerVersion.php
azjezz: psalm already can detect `array{1: string, 2: string, 3: string}|null` return type, i will move the assertion above below this call and check for null instead.

Dockerfile
Ocramius: Works for me - just unfortunate having to add this stuff to the image, heh.

src/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelog.php
Ocramius: Excellent, good to know! :+1:

src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
azjezz: done 👍

src/Github/Api/V3/CreatePullRequestThroughApiCall.php
azjezz: hm, actually this doesn't need to be set to true;

src/Github/CreateReleaseTextThroughChangelog.php
Ocramius: That's alright, was just confused by the code, but the explanation suffices :+1:

src/Gpg/ImportGpgKeyFromStringViaTemporaryFile.php
azjezz: not sure how to test it? escaping is disabled here.

Ocramius: Possibly not testable: just saw that STDERR has to be piped at process open, if you want to capture it.

Looking at https://github.com/laminas/automatic-releases/pull/138/checks?check_run_id=2638984783#step:8:11, I see no STDERR, so I suppose this doesn't need testing here.

This still brings me to another point: symfony CLI does escape shell arguments by default, what happens with this library instead? Is this the last argument of `Shell\execute()`?

azjezz: yes, Shell\execute also escapes by default, unless the last argument is set to false ( which is done here, and we escape the only argument that needs escaping ).

see https://github.com/azjezz/psl/blob/1.8.x/src/Psl/Shell/execute.php#L44

Ocramius: Excellent, thanks!

src/Github/Api/GraphQL/Query/GetMilestoneChangelog/Response/Label.php
```diff
@@ -37,17 +37,16 @@ private function __construct(
...
+     * @psalm-suppress ImpureMethodCall     {@see https://github.com/azjezz/psl/issues/130}
+     * @psalm-suppress ImpureFunctionCall   {@see https://github.com/azjezz/psl/issues/130}
      */
     public static function fromPayload(array $payload): self
     {
-        Assert::keyExists($payload, 'color');
-        Assert::keyExists($payload, 'name');
-        Assert::keyExists($payload, 'url');
-        Assert::stringNotEmpty($payload['color']);
-        Assert::regex($payload['color'], '/^[0-9a-f]{6}$/i');
```
bendavies, line 48: you've lost the regex check

azjezz: ops, fixed.

and added a test to ensure it won't be lost again.

azjezz: diff: https://github.com/laminas/automatic-releases/compare/89ecec8ce34b43e1e6bdd0f876e5f5a4dbaa4a64..6ad991b65aa5b9d52febf91dc9ca3f812a80952e

test/unit/Changelog/BumpAndCommitChangelogVersionViaKeepAChangelogTest.php
```diff
@@ -17,28 +17,23 @@
...
-use function Safe\tempnam;
-use function sprintf;
-use function sys_get_temp_dir;
-use function unlink;
 
 class BumpAndCommitChangelogVersionViaKeepAChangelogTest extends TestCase
 {
     /** @var CheckoutBranch&MockObject */
-    private $checkoutBranch;
+    private MockObject | CheckoutBranch $checkoutBranch;
```
Ocramius: These `|` still emained: need another pass on these mocks

test/unit/Github/Api/V3/CreateMilestoneTest.php
```diff
@@ -12,22 +12,18 @@
...
 class CreateMilestoneTest extends TestCase
 {
     /** @var ClientInterface&MockObject */
-    private $httpClient;
+    private ClientInterface $httpClient;
     /** @var RequestFactoryInterface&MockObject */
-    private $messageFactory;
-    /** @var LoggerInterface&MockObject */
-    private $logger;
+    private MockObject | RequestFactoryInterface $messageFactory;
```
Ocramius: Moar `MockObject |` to drop

Ocramius (changes requested):

Ocramius (approved):
Excellent refactoring, thanks!

src/Github/Api/V3/CreatePullRequestThroughApiCall.php
```diff
@@ -63,16 +62,14 @@ public function __invoke(
 
         $response = $this->client->sendRequest($request);
 
+        Psl\invariant($response->getStatusCode() >= 200 || $response->getStatusCode() <= 299, 'Failed to create pull request through GitHub API.');
```
zerkms, line 65: It should be `&&` here

/cc @Ocramius

src/Github/Api/V3/SetDefaultBranchThroughApiCall.php
```diff
@@ -49,21 +48,18 @@ public function __invoke(
 
         $request
             ->getBody()
-            ->write(json_encode(['default_branch' => $defaultBranch->name()]));
+            ->write(Json\encode(['default_branch' => $defaultBranch->name()]));
 
         $response = $this->client->sendRequest($request);
 
+        Psl\invariant($response->getStatusCode() >= 200 || $response->getStatusCode() <= 299, 'Failed to set default branch through GitHub API.');
```
zerkms, line 55: It should be `&&` here

/cc @Ocramius

src/Github/Api/GraphQL/RunGraphQLQuery.php
```diff
@@ -56,15 +54,10 @@ public function __invoke(
...
 
-        Assert::same($response->getStatusCode(), 200);
-
-        $responseData = json_decode($responseBody, true);
-
-        Assert::isMap($responseData);
-        Assert::keyNotExists($responseData, 'errors');
-        Assert::keyExists($responseData, 'data');
-        Assert::isArray($responseData['data']);
+        Type\literal_scalar(200)->assert($response->getStatusCode());
```
zerkms, line 57: `Assert::keyNotExists($responseData, 'errors');` <- this assertion was lost

/cc @Ocramius

src/Github/Api/V3/CreatePullRequestThroughApiCall.php
Ocramius: Ouch, will revert these, thanks for noticing!

src/Environment/EnvironmentVariables.php
```diff
@@ -97,9 +100,9 @@ public static function fromEnvironment(ImportGpgKeyFromString $importKey): self
      */
     private static function getenv(string $key): string
     {
-        $value = getenv($key);
+        $value = Env\get_var($key);
 
-        Assert::stringNotEmpty($value, sprintf('Could not find a value for environment variable "%s"', $key));
+        Psl\invariant($value !== null && $value !== '', Str\format('Could not find a value for environment variable "%s"', $key));
```
bendavies, line 105: @azjezz why not `Type\non_empty_string()->matches($value)` here?

azjezz: we, you can do that too, same thing

bendavies: it doesn't work actually, I found.

```
ERROR: MoreSpecificReturnType - The declared return type 'non-empty-string' for EnvironmentVariables::getenv is more specific than the inferred return type 'string' (see https://psalm.dev/070)

ERROR: LessSpecificReturnStatement -- The type 'string' is more general than the declared return type 'non-empty-string' for EnvironmentVariables::getenv (see https://psalm.dev/129)
```

bendavies: ah, yes it does. i was trying
```php
Psl\invariant(
    Type\non_empty_string()->assert($value),
    Str\format('Could not find a value for environment variable "%s"', $key)
);
```
instead of just 
```php
Type\non_empty_string()->assert($value);
```

but i guess you then lose the custom invariant error message.

azjezz: yea, you lose the message with assert, maybe we could add a costume message as a second argument, not sure 🤔

bendavies: Sounds good!

Issue #198, opened 2022-07-26
`git` broken on latest `1.14.0` image


Ocramius:
### Bug Report

`1.14.0` crashes and leaks repository secrets.

|    Q        |   A
|------------ | ------
| Version(s)  | 1.14.0

#### Summary

The `1.14.0` image has now been dropped, as a security precaution.

Two problems arise:

 1. in case of `git` operations failing, we get a crash with un-masked secrets (**really** bad, although not exploitable, luckily)
 2. because `git` added "security" around operating on repositories owned by other users, we have crashing release processes, like https://github.com/scoutapp/scout-apm-php-ext/runs/7518545888?check_suite_focus=true

Related: https://github.com/laminas/laminas-continuous-integration-action/pull/99

Ocramius:
I verified that 2 pull operations happened for `1.14.0`: one was me, one was @asgrim. No other affected users, so deleting the tag was a safe approach here.

Yes, I did delete the git tag and the docker image too.

Ocramius:
We basically need a post-startup `git config --global --add safe.directory '*'` command to be executed, I think.

Ocramius:
Probably need to use a better credential handling around the token: https://stackoverflow.com/a/61465734/347063

That's something that we just need to test somehow 🤔

weierophinney:
I've seen #199 - will you be addressing the `safe.directory` setting in a separate PR?

Ocramius:
I hope so: needed to go AFK in a hurry, so I couldn't finish what I started

Ocramius:
Handled in #200

PR #199, opened 2022-07-26
Use `git config credential.helper 'store --file=/tmp/file' for `git push`

Ocramius:
This should fix some underlying design issues described in #198
    
The full URI of the repository, which includes a token in plaintext, is now no
longer the actual `git remote` URI, and is instead stored in a temporary
file which `git` uses to authenticate with the remote.
    
This prevents the remote URI from leaking in `STDOUT`/`STDERR`, which are
visible in case of crashes.

This is a **partial** fix of #198 - the issue with `git` being broken upfront is still there.

### DIFF
--- a/src/Application/Command/BumpChangelogForReleaseBranch.php
+++ b/src/Application/Command/BumpChangelogForReleaseBranch.php
@@ -44,12 +44,15 @@ final class BumpChangelogForReleaseBranch extends Command
     {
         $milestoneClosedEvent = ($this->loadEvent)();
         $repositoryName       = $milestoneClosedEvent->repository();
-        $repositoryCloneUri   = $repositoryName->uriWithTokenAuthentication($this->environment->githubToken());
         $repositoryPath       = $this->environment->githubWorkspacePath();
 
         Psl\invariant(Filesystem\is_directory($repositoryPath . '/.git'), 'Workspace is not a GIT repository.');
 
-        ($this->fetch)($repositoryCloneUri, $repositoryPath);
+        ($this->fetch)(
+            $repositoryName->uri(),
+            $repositoryName->uriWithTokenAuthentication($this->environment->githubToken()),
+            $repositoryPath
+        );
 
         $mergeCandidates = ($this->getMergeTargets)($repositoryPath);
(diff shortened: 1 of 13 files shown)


src/Application/Command/BumpChangelogForReleaseBranch.php
```diff
@@ -44,12 +44,15 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
         $repositoryPath       = $this->environment->githubWorkspacePath();
 
         Psl\invariant(Filesystem\is_directory($repositoryPath . '/.git'), 'Workspace is not a GIT repository.');
 
-        ($this->fetch)($repositoryCloneUri, $repositoryPath);
+        ($this->fetch)(
+            $repositoryName->uri(),
+            $repositoryName->uriWithTokenAuthentication($this->environment->githubToken()),
+            $repositoryPath
+        );
```
weierophinney, line 55: Trying to understand how this fixes the problem - is it because in the case of a crash, only the first argument (here, `$repositoryName->uri()`, previously, a version of the URI with the token included) is ever logged?

Ocramius: It was in the exception message.

I think a validating test case is needed.

test/unit/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemoteTest.php
```diff
@@ -105,4 +106,27 @@ public function testFetchesAndSetsCurrentUserWithoutOrigin(): void
...
+        Filesystem\delete_file($notAGitDirectory);
+        Filesystem\create_directory($notAGitDirectory);
+
+        try {
+            (new FetchAndSetCurrentUserByReplacingCurrentOriginRemote($this->variables))
+                ->__invoke($sourceUri, $sourceUriWithSecrets, $notAGitDirectory);
+        } catch (Throwable $failure) {
+            self::assertDoesNotMatchRegularExpression('/SUPERSECRET/m', $failure->getMessage());
+        }
+    }
```
Ocramius, line 131: Verified locally that this test fails on `1.14.x`, while succeeds on this branch.

PR #200, opened 2022-07-26
Fix #198 - make GIT's `safe.directory` `'*'` by default on repository fetch

### LINKED ISSUE #198: `git` broken on latest `1.14.0` image

Ocramius:
This is an ugly side-effect, and it will modify the environment of users that rely on this package
outside the context of the delivered docker container, but it is a good workaround for the security
and compatibility mess that GIT 2.35.3 introduced.

In fact, this application was never meant to run in multi-user mode, and the container it runs will
not be affected by side effects (mostly git hooks) in freshly checked out code.

Ref: https://stackoverflow.com/a/71904131/347063
Ref: https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory
Ref: https://marc.info/?l=git&m=164989570902912&w=2
Ref: https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9
Ref: https://github.com/git-for-windows/git/issues/3809#issuecomment-1102681740

### DIFF
--- a/src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
+++ b/src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
@@ -31,6 +31,7 @@ final class FetchAndSetCurrentUserByReplacingCurrentOriginRemote implements Fetc
 
         $credentialStore = Filesystem\create_temporary_file();
 
+        Shell\execute('git', ['config', '--global', '--add', 'safe.directory', '*'], $repositoryRootDirectory);
         Shell\execute('git', ['config', 'credential.helper', 'store --file=' . $credentialStore], $repositoryRootDirectory);
         File\write($credentialStore, $uriWithCredentials->__toString());
         Shell\execute('git', ['remote', 'add', 'origin', $repositoryUri->__toString()], $repositoryRootDirectory);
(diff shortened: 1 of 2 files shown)


src/Git/FetchAndSetCurrentUserByReplacingCurrentOriginRemote.php
```diff
@@ -31,6 +31,7 @@ public function __invoke(
 
         $credentialStore = Filesystem\create_temporary_file();
 
+        Shell\execute('git', ['config', '--global', '--add', 'safe.directory', '*'], $repositoryRootDirectory);
```
Ocramius, line 34: Ugly but necessary :-\

boesing (approved):
LGTM.
The test looks a bit weird as it supposed to be an integration test since you are validating that the command was executed but I don't want to be too picky here.

Ocramius:
:ship: 

Trying out a release ASAP.

PR #204, opened 2022-08-17
Check the response code is `>= 200 && <= 299` on release

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

I don't see an issue for it, but the response code check is an `OR` when it should be an `AND`.

### DIFF
--- a/src/Github/Api/V3/CreateReleaseThroughApiCall.php
+++ b/src/Github/Api/V3/CreateReleaseThroughApiCall.php
@@ -60,7 +60,7 @@ final class CreateReleaseThroughApiCall implements CreateRelease
 
         $response = $this->client->sendRequest($request);
 
-        Psl\invariant($response->getStatusCode() >= 200 || $response->getStatusCode() <= 299, 'Failed to create release through GitHub API.');
+        Psl\invariant($response->getStatusCode() >= 200 && $response->getStatusCode() <= 299, 'Failed to create release through GitHub API.');
 
         $responseBody = $response
             ->getBody()
--- a/test/unit/Github/Api/V3/CreateReleaseThroughApiCallTest.php
+++ b/test/unit/Github/Api/V3/CreateReleaseThroughApiCallTest.php
@@ -1,0 +1,186 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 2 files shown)


src/Github/Api/V3/CreateReleaseThroughApiCall.php
```diff
@@ -60,7 +60,7 @@ public function __invoke(
 
         $response = $this->client->sendRequest($request);
 
-        Psl\invariant($response->getStatusCode() >= 200 || $response->getStatusCode() <= 299, 'Failed to create release through GitHub API.');
+        Psl\invariant($response->getStatusCode() >= 200 && $response->getStatusCode() <= 299, 'Failed to create release through GitHub API.');
```
Ocramius, line 63: Fine by me, but we didn't catch this during refactoring due to missing tests: we can haz?

Ocramius (changes requested):

src/Github/Api/V3/CreateReleaseThroughApiCall.php
internalsystemerror: u now haz 😊

test/unit/Github/Api/V3/CreateReleaseThroughApiCallTest.php
```diff
@@ -0,0 +1,186 @@
...
+use PHPUnit\Framework\TestCase;
+use Psl\Exception\InvariantViolationException;
+use Psl\Json\Exception\DecodeException;
+use Psl\SecureRandom;
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\RequestInterface;
+
+/** @covers \Laminas\AutomaticReleases\Github\Api\V3\CreateReleaseThroughApiCall */
+final class CreateReleaseThroughApiCallTest extends TestCase
```
Ocramius, line 22: Whaaaaaat, this class was completely untested? 😱 

Thanks for adding this, @internalsystemerror!

Ocramius (approved):

PR #211, opened 2022-09-19
Log HTTP requests/responses (only non-sensitive data)

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

This patch introduces general logging for:

 * HTTP requests sent to GitHub
 * HTTP responses received from GitHub

This can become a bit verbose, but verbosity could be adjusted later on, if needed: for now, we need this to figure out when/if things go wrong during a release, which is more important than clean output.

### DIFF
--- a/bin/console.php
+++ b/bin/console.php
@@ -37,6 +37,9 @@ use Laminas\AutomaticReleases\Github\Event\Factory\LoadCurrentGithubEventFromGit
 use Laminas\AutomaticReleases\Github\JwageGenerateChangelog;
 use Laminas\AutomaticReleases\Github\MergeMultipleReleaseNotes;
 use Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromStringViaTemporaryFile;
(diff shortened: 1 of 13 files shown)


cspray:
LGTM :+1:

Ocramius:
:ship: as per @cspray's feedback.

I applied some coverage improvements too, while I was at it :)

Ocramius:
Hmm, the logger doesn't seem to be in use:

```

In AssertException.php line 32:
                                        
  [Psl\Type\Exception\AssertException]  
  Expected "200", got "int".  
```

Ref: https://github.com/Roave/DocbookTool/actions/runs/3084714475/jobs/4987180006#step:4:13

Seems like the logger output doesn't make it to GitHub's STDERR at all?

Ocramius:
Ah, `LOG_LEVEL=DEBUG` should be used there

PR #220, opened 2022-12-16
Enable debug mode for GitHub Actions

ghostwriter:
- Check if the environment variable `ACTIONS_RUNNER_DEBUG ` is present, to set the log level to `DEBUG` mode via GitHub Actions.
- Note: it does not overwrite the default `LOG_LEVEL` environment variable.

### DIFF
--- a/src/Environment/EnvironmentVariables.php
+++ b/src/Environment/EnvironmentVariables.php
@@ -6,10 +6,11 @@ namespace Laminas\AutomaticReleases\Environment;
 
 use Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromString;
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
-use Psl;
-use Psl\Env;
-use Psl\Iter;
-use Psl\Str;
+
+use function Psl\Env\get_var;
+use function Psl\invariant;
+use function Psl\Iter\contains;
+use function Psl\Str\format;
 
 /** @psalm-immutable */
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Very clever usage of the `debug` flag of github actions: nice!

boesing:
I'd say we should enable debug logging in every case.
What I experienced many times now was, that the action creates the `git tag`  and then fails to create the release on github due to the API is not providing the expected status code.

If you then "re-run" the action again with the debug flag, `git tag` command will fail and you only get debug up to the `git tag` error but you are unable to understand the invalid API response error.

Ocramius:
I'd be up for it: was hesitant because I'm worried about secret leaks in logs.

I already hardened the logger against it:

https://github.com/laminas/automatic-releases/blob/49d9580079916c72cebd068aa07852dc9b548eeb/src/Monolog/ConvertLogContextHttpRequestsIntoStrings.php#L35-L40

If this filtering is sufficient for our future use-cases, then we should be OK with having verbose logging at all times.

Ocramius:
Another thing to think about is if we can re-design the default pipeline so it is in "stages" that can be retried individually :thinking:

boesing:
I guess that would be the best. Or at least having checks like "does tag already exist then assume tagging was successfully"?
If a tag is available thats most probably because something went wrong earlier OR some1 failed setting up his project.
Maybe double-checking if the current commit is the same commit which is already tagged or sth. like this.

PR #228, opened 2022-12-21
Apply PHP 8.2 syntax

samsonasik:
|    Q          |   A
|-------------- | ------
| QA            | yes 

### Description

Since composer.json require php 8.2, the syntax can be updated with php 8.2 feature.


samsonasik:
All green 🎉

Ocramius (approved):
Looking good, thanks @samsonasik!

PR #230, opened 2023-01-01
Skip creating GIT tag if one already exists, enabling re-running of the action if a release creation fails

### LINKED ISSUE #222: [RFC]Re-running automatic-releases actions after failure
|    Q        |   A
|------------ | ------
| Version(s)  | 1.20.x

Check if a `git tag` exists that matches the `milestone` version and `commit sha`, to then skip the "tagging" step.

#### Current behavior

GitHub Action fails due to `git tags` created in a previous run.

#### How to reproduce

`re-run` the GitHub Action after failing to publish a release due to a GitHub API error.

#### Expected behavior

No errors.

ghostwriter:
Resolves #222

### DIFF
--- a/src/Git/CreateTagViaConsole.php
+++ b/src/Git/CreateTagViaConsole.php
@@ -10,13 +10,16 @@ use Psl\Env;
 use Psl\File;
 use Psl\Filesystem;
 use Psl\Shell;
+use Psr\Log\LoggerInterface;
 
 use function sprintf;
 
 final class CreateTagViaConsole implements CreateTag
 {
-    public function __construct(private HasTag $hasTag)
-    {
+    public function __construct(
+        private readonly HasTag $hasTag,
+        private readonly LoggerInterface $logger,
+    ) {
     }
 
     public function __invoke(
@@ -26,7 +29,11 @@ final class CreateTagViaConsole implements CreateTag
         string $changelog,
         SecretKeyId $keyId,
     ): void {
-        if (($this->hasTag)($repositoryDirectory, $tagName) === true) {
+        if (($this->hasTag)($repositoryDirectory, $tagName)) {
+            $this->logger->info(
+                sprintf('[CreateTagViaConsole] Skipping this step; tag "%s" already exists.', $tagName),
+            );
+
             return;
         }
 
--- a/src/Git/CreateTagViaConsole.php
+++ b/src/Git/CreateTagViaConsole.php
@@ -6,13 +6,18 @@ namespace Laminas\AutomaticReleases\Git;
 
 use Laminas\AutomaticReleases\Git\Value\BranchName;
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
-use Psl\Env;
-use Psl\File;
-use Psl\Filesystem;
-use Psl\Shell;
+
+use function Psl\Env\temp_dir;
+use function Psl\File\write;
+use function Psl\Filesystem\create_temporary_file;
+use function Psl\Shell\execute;
 
 final class CreateTagViaConsole implements CreateTag
 {
+    public function __construct(private HasTag $hasTag)
+    {
+    }
+
     public function __invoke(
         string $repositoryDirectory,
         BranchName $sourceBranch,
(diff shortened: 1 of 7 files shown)


src/Git/CreateTagViaConsole.php
```diff
@@ -6,25 +6,39 @@
...
         string $repositoryDirectory,
         BranchName $sourceBranch,
         string $tagName,
         string $changelog,
         SecretKeyId $keyId,
     ): void {
-        $tagFileName = Filesystem\create_temporary_file(Env\temp_dir(), 'created_tag');
+        if (($this->hasTag)($repositoryDirectory, $tagName) === true) {
+            return;
+        }
```
Ocramius: Worth logging a warning here

src/Git/CreateTagViaConsole.php
```diff
@@ -6,25 +6,39 @@
...
+
     public function __invoke(
         string $repositoryDirectory,
         BranchName $sourceBranch,
         string $tagName,
         string $changelog,
         SecretKeyId $keyId,
     ): void {
-        $tagFileName = Filesystem\create_temporary_file(Env\temp_dir(), 'created_tag');
+        if (($this->hasTag)($repositoryDirectory, $tagName) === true) {
```
Ocramius: `=== true` to be omitted: the interface states it's a boolean return value anyway

src/Git/CreateTagViaConsole.php
```diff
@@ -6,25 +6,39 @@
...
 use Laminas\AutomaticReleases\Gpg\SecretKeyId;
-use Psl\Env;
-use Psl\File;
-use Psl\Filesystem;
-use Psl\Shell;
+
+use function Psl\Env\temp_dir;
+use function Psl\File\write;
+use function Psl\Filesystem\create_temporary_file;
+use function Psl\Shell\execute;
```
Ocramius: Noticed this on other patches of yours: probably done by the IDE, but consider not importing single functions, but just their namespace.

It is much much easier to identify `Shell\execute()` in the middle of the code, rather than `execute()`. This becomes much more visible when working with `Psl\Type\*`

src/Git/CreateTagViaConsole.php
```diff
@@ -6,25 +6,39 @@
...
-        File\write($tagFileName, $changelog);
+        execute('git', ['checkout', $sourceBranch->name()], $repositoryDirectory);
 
-        Shell\execute('git', ['checkout', $sourceBranch->name()], $repositoryDirectory);
-        Shell\execute('git', ['tag', $tagName, '-F', $tagFileName, '--cleanup=whitespace', '--local-user=' . $keyId->id()], $repositoryDirectory);
+        execute(
+            'git',
+            ['tag', $tagName, '-F', $tagFileName, '--cleanup=whitespace', '--local-user=' . $keyId->id()],
+            $repositoryDirectory,
+        );
```
Ocramius: I think all this stays mostly unchanged, if the import changes are reverted/squashed

src/Git/HasTag.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\AutomaticReleases\Git;
+
+interface HasTag
+{
+    /** @param non-empty-string $repositoryDirectory */
```
Ocramius: `$tagName` should probably also be a `non-empty-string`

src/Git/HasTagViaConsole.php
```diff
@@ -0,0 +1,23 @@
...
+
+use function Psl\Shell\execute;
+use function str_contains;
+use function trim;
+
+final class HasTagViaConsole implements HasTag
+{
+    public function __invoke(string $repositoryDirectory, string $tagName): bool
+    {
+        $outout = execute('git', ['tag', '--list', $tagName], $repositoryDirectory);
```
Ocramius: `$outout` probably a typo

src/Git/HasTagViaConsole.php
```diff
@@ -0,0 +1,23 @@
...
+
+        if (trim($outout) === '') {
+            return false;
+        }
+
+        return str_contains($outout, $tagName);
```
Ocramius: Does `git` not provide an exit code for stuff it doesn't find? 🤔

Ocramius (commented):
Overall good direction.

Just to note here that this will get us un-stuck if a tag is created, but a release isn't: we still have many such possible roadblocks with merge-up PRs, new release milestone creation and default tag switching.

Not sure if those are relevant for now: this one is a very good start already 💪

src/Git/HasTagViaConsole.php
ghostwriter: "No", or at least not to my knowledge.

https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---list
https://stackoverflow.com/questions/4917871/does-git-return-specific-return-error-codes

Ocramius (approved):
Thanks @ghostwriter!

PR #247, opened 2023-09-13
Ci fixes

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

CI is failing in various ways because of some trivial problems…

- Fix linting issues with README.md
- Bump all deps and prevent composer installation problems by denying the `php-http/discovery` composer plugin
- Fix deprecations in PHPUnit and `php-http/discovery`

### DIFF
--- a/infection.json.dist
+++ b/infection.json.dist
@@ -12,6 +12,6 @@
     "mutators": {
         "@default": true
     },
-    "minMsi": 98,
-    "minCoveredMsi": 100
+    "minMsi": 97.4,
+    "minCoveredMsi": 98
 }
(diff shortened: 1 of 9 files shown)


gsteel:
@Ocramius any ideas on improving MSI here? Most of the mutants are escaping via calls to `Shell\Execute` and it's not like that can be mocked 🤷‍♂️

infection.json.dist
```diff
@@ -12,6 +12,6 @@
     "mutators": {
         "@default": true
     },
-    "minMsi": 98,
-    "minCoveredMsi": 100
+    "minMsi": 97.4,
+    "minCoveredMsi": 98
```
Ocramius, line 16: This dipping below 100% worries me a bit :D

Ocramius: > https://github.com/laminas/automatic-releases/pull/247#issuecomment-1717298203

I'd say that we can leave it be then - sad, but understandable.

Ocramius (approved):
Thanks @gsteel!

PR #282, opened 2026-02-11
Fix creating minor release if new major release branch already exist

DanielBadura:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

This patch fixes the minor branch creation if already a new major branch exist. This is useful if a new major release is already in the work but new feature are still coming into the current latest major.

The case is as follows:
When we already have these branches: 1.0.x, 1.1.x, 1.2.x, 2.0.x And we release 1.2.0, then we expect that 1.3.x is created and set as a new default branch.

### DIFF
--- a/feature/automated-releases.feature
+++ b/feature/automated-releases.feature
@@ -38,7 +38,7 @@ Feature: Automated releases
     When I close milestone "2.0.0"
     Then tag "2.0.0" should have been created on branch "2.0.x"
 
-  Scenario: If a new major release branch exists, the tool does not create a new minor release
+  Scenario: If matching minor release branch exists, the tool does not create a new minor release
     Given following existing branches:
       | name    |
       | 1.0.x   |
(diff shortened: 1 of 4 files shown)


gsteel (commented):
This LGTM, but I'm not overly familiar with this codebase so I'll defer to perhaps @Ocramius for review :)

Xerkus:
For our purposes new minor branch needs to be created only if `2.0.x` exist but `2.1.x` does not. In other words there were no 2.0.0 release yet.

DanielBadura:
So is this a rule you want to enforce here? This also aligns with our purpose so I'm fine with that.
As of SemVer I think it should be fine releasing 1.23.0 even if 2.3.0 is already out but i guess up-merges would be tricky.. 😅

Ocramius (approved):
Let's give this a shot, thanks @DanielBadura!

PR #305, opened 2026-09-03
Set GNUPGHOME to /tmp in the release image

### LINKED ISSUE #304: GPG key import fails with keyboxd SQLite lock timeout in the refreshed 1.27.0 image
Since the `ghcr.io/laminas/automatic-releases:1` image was refreshed to 1.27.0 (published 2026-09-01, after being stuck on 1.24.0, see #301), the GPG key import fails intermittently inside the container.

Failing job: https://github.com/doctrine/mongodb-odm/actions/runs/33754680070/job/100646127583

In that run the `laminas:automatic-releases:release` step succeeded, then the very next step, `laminas:automatic-releases:create-merge-up-pull-request`, crashed on the same key import. So the tag and the GitHub release were created, but the merge-up PR, the default branch switch and the milestone creation were all lost. We had to redo them by hand.

```
Fatal error: Uncaught Psl\Shell\Exception\FailedExecutionException: Shell command "exec gpg '--import' '/tmp/imported-keySAmxcZ3L'" returned an exit code of "2".

STDERR:
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: keydb_get_keyblock failed: Value not found
    gpg: error writing keyring '[keyboxd]': SQL library used incorrectly
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: keydb_search failed: Operation timed out
    gpg: key A1A906E088C3CEE4: failed to re-lookup public key: Operation timed out
    gpg: error reading '/tmp/imported-keySAmxcZ3L': Operation timed out
    gpg: import from '/tmp/imported-keySAmxcZ3L' failed: Operation timed out
    gpg: Total number processed: 0
    gpg:       secret keys read: 1
 in /app/vendor/azjezz/psl/src/Psl/Shell/execute.php:185
Stack trace:
#0 /app/src/Gpg/ImportGpgKeyFromStringViaTemporaryFile.php(22)
#1 /app/src/Environment/EnvironmentVariables.php(66)
#2 /app/bin/console.php(67)
```

The new base image ships a GPG version that stores public keys in the `keyboxd` SQLite database instead of a plain keyring. A `keyboxd` daemon left over from an earlier step in the same job holds the lock, and the import times out.

Two things that would help:

1. Disable `keyboxd` in the image, for example by writing `use-keyboxd` out of `common.conf` or by setting `--no-use-keyboxd`, so key import goes back to the plain keyring and no daemon is involved.
2. Failing that, kill any running `keyboxd`/`gpg-agent` before the import, or use a per-invocation `GNUPGHOME`, so consecutive steps in a job do not fight over the same database.

Independently, a run that partially succeeds is expensive to recover from. Making the release steps idempotent, or at least documenting the manual recovery, would reduce the damage when this happens.

GromNaN:
In GitHub Actions, `$HOME` is a volume shared by every step of a job, so the default GnuPG home (`$HOME/.gnupg`) leaks the imported private key to later steps, and GnuPG 2.4 leaves a `keyboxd` database and a dead socket that make the next import time out.

Setting `GNUPGHOME=/tmp` keeps the GnuPG state inside the container. `/tmp` is container-local, already exists, and stays writable under a read-only root filesystem, so no directory needs to be created.

Fixes #304

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -18,4 +18,19 @@ jobs:
         uses: "actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1" # v7
 
       - name: "Docker Build"
-        run: "DOCKER_BUILDKIT=1 docker build ."
+        run: "DOCKER_BUILDKIT=1 docker build --tag 'automatic-releases:ci' ."
+
+      # Regression test for https://github.com/laminas/automatic-releases/issues/304: a job runs this image
+      # several times with the same $HOME volume, and the second GPG import used to time out on the state
+      # left behind by the first one.
+      - name: "Import a GPG key twice, as a release job does"
+        run: |
+          mkdir -p "${RUNNER_TEMP}/github-home"
+          for attempt in 1 2; do
+            timeout 120 docker run --rm \
+              --volume "${RUNNER_TEMP}/github-home:/github/home" \
+              --volume "${GITHUB_WORKSPACE}/test/asset:/asset:ro" \
+              --env "HOME=/github/home" \
+              --entrypoint gpg \
+              'automatic-releases:ci' --batch --import /asset/dummy-gpg-key.asc
+          done
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,15 @@ COPY src /app/src/
 
 RUN composer dump-autoload -a --no-dev
 
+# Keep the GnuPG state inside the container. In GitHub Actions $HOME is a volume shared by every step of a
+# job, so a GNUPGHOME under $HOME leaks from one step to the next: the imported private key stays readable
+# by later steps, and GnuPG 2.4 makes it worse by enabling keyboxd when it creates a default home directory,
+# leaving a database and a dead socket that make the next import time out. A directory created here is fresh
+# in every container, so no state is ever shared. As a bonus, GnuPG 2.4 leaves an existing GNUPGHOME on the
+# plain keyring instead of switching it to keyboxd.
+ENV GNUPGHOME=/gnupg
+RUN mkdir -m 0700 /gnupg
+
 ENV SHELL_VERBOSITY=3
 
 ENTRYPOINT ["/app/bin/console.php"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,9 @@ COPY src /app/src/
 
 RUN composer dump-autoload -a --no-dev
 
+# Keep the GnuPG state inside the container so it does not leak through the shared $HOME volume.
+ENV GNUPGHOME=/tmp
+
 ENV SHELL_VERBOSITY=3
 
 ENTRYPOINT ["/app/bin/console.php"]


Ocramius:
@gsteel seems reasonable to me

Ocramius:
Hmm, doesn't work with immutable filesystem: perhaps we should just use an explicit temporary directory?

gsteel:
@Ocramius - do you mean if the container is run with `--read-only`? or the host filesystem is immutable?

GromNaN:
@Ocramius You were right, baking the directory into the image does not work with a read-only root filesystem. The fix is now a single line: `ENV GNUPGHOME=/tmp` in the Dockerfile.

`/tmp` is the explicit temporary directory: it is container-local, so the imported key never lands on the `$HOME` volume shared between steps; it already exists, so no directory needs to be created at build or runtime, which is what broke under an immutable filesystem; and GnuPG 2.4 keeps it on the plain keyring (`pubring.kbx`) instead of `keyboxd` because the directory already exists.

`HOME` itself cannot be relocated: GitHub Actions pins `HOME=/github/home` for Docker container actions and overrides any `ENV HOME` from the Dockerfile, so `GNUPGHOME` is the lever that moves only the GnuPG state off the shared volume while leaving git's global config where it is.

I also considered moving the GnuPG home into the application, a `GnupgHome` helper injecting `GNUPGHOME` into every `gpg`/`git` `Shell\execute` call (see commit 0cdc016). It works and was verified under `--read-only --tmpfs /tmp`, but it touches four source files plus tests for no benefit over the single `ENV` line, so I dropped it.)

Ocramius (approved):
🚢 

Also fixes the release process, from what I can see.

Ocramius:
Still getting weird locking issues with this ( https://github.com/laminas/automatic-releases/actions/runs/34402108229/job/102636232702 ) 


```
Fatal error: Uncaught Psl\Shell\Exception\FailedExecutionException: Shell command "exec gpg '--import' '/tmp/imported-keyQswnRdl0'" returned an exit code of "2".

STDOUT:
    

STDERR:
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: keydb_get_keyblock failed: Value not found
    gpg: error writing keyring '[keyboxd]': SQL library used incorrectly
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: Note: database_open 134217901 waiting for lock (held by 9) ...
    gpg: keydb_search failed: Operation timed out
    gpg: key 10B7743EC1CCD759: failed to re-lookup public key: Operation timed out
    gpg: error reading '/tmp/imported-keyQswnRdl0': Operation timed out
    gpg: import from '/tmp/imported-keyQswnRdl0' failed: Operation timed out
    gpg: Total number processed: 0
    gpg:       secret keys read: 1
     in /app/vendor/azjezz/psl/src/Psl/Shell/execute.php:185
Stack trace:
#0 /app/src/Gpg/ImportGpgKeyFromStringViaTemporaryFile.php(22): Psl\Shell\execute('gpg', Array, '/github/workspa...', Array, Object(Psl\Shell\ErrorOutputBehavior))
#1 /app/src/Environment/EnvironmentVariables.php(66): Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromStringViaTemporaryFile->__invoke('-----BEGIN PGP ...')
#2 /app/bin/console.php(67): Laminas\AutomaticReleases\Environment\EnvironmentVariables::fromEnvironment(Object(Laminas\AutomaticReleases\Gpg\ImportGpgKeyFromStringViaTemporaryFile))
#3 /app/bin/console.php(181): Laminas\AutomaticReleases\WebApplication\{closure}()
#4 {main}
  thrown in /app/vendor/azjezz/psl/src/Psl/Shell/execute.php on line 185
```

Ocramius:
> @Ocramius - do you mean if the container is run with `--read-only`? or the host filesystem is immutable?

forgot to answer: indeed `--read-only` :+1:

GromNaN:
@Ocramius The locking error in that run is not a regression of this change: the run pulled the image built from the **1.27.0** release, which does not yet contain the `GNUPGHOME=/tmp` fix.

The `:1` tag is a floating tag, rebuilt only when a release is published. The 1.28.0 release was published about two minutes after that run started, so the run still used the previous `:1` image (1.27.0, default `$HOME/.gnupg`, keyboxd). The first step (release) imported the key and succeeded; the second step (merge-up), in a new container sharing the same `$HOME` volume, hit the leftover keyboxd socket and timed out, exactly as in #304.

The fix shipped here is correct and verified: with `GNUPGHOME=/tmp`, gpg uses the file-based `pubring.kbx` (no keyboxd), and `/tmp` is container-local, so nothing leaks through the shared `$HOME` volume. The current `:1` image (rebuilt from 1.28.0) already includes it.

To stop the floating `:1` tag from causing this kind of stale-image surprise again, I opened #306 to pin the image by digest and have the build workflow keep the pin up to date on each release.)

